PR #25, opened 2024-08-16
Streaming

CharlesDuboisSAP:
## Context

AI/ai-sdk-java-backlog#59.

Streaming allows the user to get an initial response from the LLM without waiting for the whole response.
This is a proof of concept

### Feature scope:
 
- [ ] PoC

## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~


foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -129,6 +144,16 @@ private <T> T execute(
     return executeRequest(request, responseType);
   }
 
+  @Nonnull
+  private <T> OpenAiChatCompletionStream<T> stream(
+      @Nonnull final String path,
+      @Nonnull final Object payload,
+      @Nonnull final Class<T> responseType) {
+    final var request = new HttpPost(path);
```
newtork, line 152: (Comment)

Not sure, whether "stream" should automatically infer "POST" http method.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
```diff
@@ -0,0 +1,82 @@
...
+   * @param response The response to process
+   * @return A {@link OpenAiChatCompletionStream} of a model class instantiated from the response
+   * @author stippi
+   */
+  private OpenAiChatCompletionStream<T> parseResponse(@Nonnull final ClassicHttpResponse response)
+      throws OpenAiClientException {
+
+    InputStream inputStream;
+    try {
+      inputStream = response.getEntity().getContent();
```
newtork, line 51: (Minor)

Needs check for `getEntity()==null`

e2e-test-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java
```diff
@@ -38,6 +46,47 @@ public static OpenAiChatCompletionOutput chatCompletion() {
     return OpenAiClient.forModel(GPT_35_TURBO).chatCompletion(request);
   }
 
+  /**
+   * Stream chat request to OpenAI
+   *
+   * @return the emitter that streams the assistant message response
+   */
+  @GetMapping("/stream")
```
newtork, line 54: (Minor)

Not a good endpoint name.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/model/OpenAiChatCompletionOutput.java
```diff
@@ -25,4 +26,26 @@ public class OpenAiChatCompletionOutput extends OpenAiCompletionOutput {
   @JsonProperty("system_fingerprint")
   @Getter(onMethod_ = @Nonnull)
   private String systemFingerprint;
+
+  public void addDelta(OpenAiChatCompletionOutput delta) {
```
newtork, line 30: (Major)

What do you need this method for? It is unused or is it not? Certainly missing JavaDoc and there's still a TODO. I assume this is a left-over(?)

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/model/OpenAiChatCompletionStream.java
```diff
@@ -0,0 +1,28 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.Setter;
+import lombok.experimental.Accessors;
+
+@Getter
+@RequiredArgsConstructor(access = AccessLevel.NONE)
+@NoArgsConstructor
+@Setter
+@Accessors(chain = true)
+public class OpenAiChatCompletionStream<T> implements AutoCloseable {
```
newtork, line 18: (Comment/Major/Discussion)

I think this class is an unnecessary wrapper. If we just exposed `Stream` we wouldn't need to care about `AutoCloseable` ourselves.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -105,6 +107,20 @@ public OpenAiChatCompletionOutput chatCompletion(
...
 
+  /**
+   * Generate a completion for the given prompt.
+   *
+   * @param parameters the prompt, including messages and other parameters.
+   * @return the completion output
+   * @throws OpenAiClientException if the request fails
+   */
+  @Nonnull
+  public OpenAiChatCompletionStream<OpenAiStreamOutput> stream(
```
newtork, line 118: (Comment/Major/Discussion)

This public method name is ambiguous:
* What does "stream" mean? Is it related to OpenAPI "stream" parameter or Java `Stream`?
* What entity is being "streamed"?

These question don't need to be answered here, but I think they should be apparent for the users of the API.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientException.java
```diff
@@ -5,8 +5,23 @@
 
 /** Generic exception for errors occurring when using OpenAI foundation models. */
 public class OpenAiClientException extends RuntimeException {
+  static final String BASE_ERROR_MESSAGE = "Request to OpenAI model failed";
   @Serial private static final long serialVersionUID = -7345541120979974432L;
 
+  /** Create a new exception with the base message: {@code Request to OpenAI model failed} */
+  public OpenAiClientException() {
```
newtork, line 12: (Major/Discussion)

Please don't allow for lazy shortcuts. Every instance where exceptions are thrown, it should have a distinct error message (and cause) such that troubleshooting remains easy.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientException.java
```diff
@@ -5,8 +5,23 @@
...
+    super(BASE_ERROR_MESSAGE);
+  }
+
+  /**
+   * Create a new exception with the base message: {@code Request to OpenAI model failed}
+   *
+   * @param e the cause
+   */
+  public OpenAiClientException(@Nonnull final Exception e) {
+    super(BASE_ERROR_MESSAGE, e);
```
newtork, line 22: (Comment)

Same issue as above, but a second thing: Please always prefer constructor chaining (`this(..., ...)`) over calling `super(..., ...)` directly.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
CharlesDuboisSAP: The regular `execute` method above also automatically infers HttpPost

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiResponseHandler.java
```diff
@@ -85,19 +94,28 @@ private void buildExceptionAndThrow(@Nonnull final ClassicHttpResponse response)
...
     if (error == null) {
-      throw exception;
+      throw baseException;
     }
-    final var message =
-        "Request to OpenAI model failed with %s %s and error message: '%s'"
-            .formatted(response.getCode(), response.getReasonPhrase(), error.getMessage());
-    throw new OpenAiClientException(message);
+    throw new OpenAiClientException(
+        baseException.getMessage() + "and error message: '%s'".formatted(error.getMessage()));
```
newtork, line 119: (Comment/Question)

This exception message recycling doesn't look like good practice to me, have we done that before? Or is this explained somewhere else as good practice?

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
CharlesDuboisSAP: Changed to `streamChatCompletion()`

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiResponseHandler.java
CharlesDuboisSAP: Both answers: no and no.
But this is how it worked before: creating 2 exceptions and reusing the message of the first to create the second.

e2e-test-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java
```diff
@@ -38,6 +46,50 @@ public static OpenAiChatCompletionOutput chatCompletion() {
...
+                            log.error(Arrays.toString(e.getStackTrace()));
+                            emitter.completeWithError(e);
+                          }
+                        });
+              } finally {
+                emitter.complete();
+              }
+            });
+    // MediaType.TEXT_EVENT_STREAM allows the browser to display the content as it is streamed
+    return ResponseEntity.ok().contentType(MediaType.TEXT_EVENT_STREAM).body(emitter);
```
MatKuhr, line 90: this is a problem, because it always returns 200, even when `streamChatCompletion(request)` fails immediately, e.g. with a 400 bad request.

Instead, `ResponseEntity.ok()` should only be invoked, once `streamChatCompletion` got a 200 (and potentially started to get some content). Not sure how to implement this exactly, maybe we can just move the `streamChatCompletion` call out of the thread, and only pass the stream to the thread? But maybe that is awkward with the try-with-resource/finally setup?

CharlesDuboisSAP: 400 is throwing `OpenAiStreamingHandler:39`, which will be treated by our `ControllerExceptionHandler`

MatKuhr: > which will be treated by our ControllerExceptionHandler

I am afraid it won't, because it's caught by the try inside the thread and has no impact on the `ok()` response.

CharlesDuboisSAP: The try doesn't catch anything, it's a try-with-resources to close the stream.

MatKuhr: true, the try doesn't catch it, it only closes the stream. But because it's inside the thread the exception will not be thrown in the main thread, so yes moving it should fix it

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -105,6 +108,41 @@ public OpenAiChatCompletionOutput chatCompletion(
...
+  public Stream<String> simpleStreamChatCompletion(
+      @Nonnull final OpenAiChatCompletionParameters parameters) throws OpenAiClientException {
+    return streamChatCompletion(parameters)
+        .peek(
+            delta -> {
+              final String finishReason = delta.getFinishReason();
+              if (finishReason != null && finishReason.equals("content_filter")) {
+                throw new OpenAiClientException("Content filter filtered the output.");
+              }
+            })
```
MatKuhr, line 142: (minor) maybe extracting this into a small method is a bit more readable? E.g.:
```suggestion
        .peek(this::throwOnContentFilter)
```

README.md
```diff
@@ -197,13 +197,93 @@ final String resultMessage = result.getChoices().get(0).getMessage().getContent(
...
-### Chat completion with a model not in defined `OpenAiModel`
+### Chat completion with a model not defined in `OpenAiModel`
 
 ```java
 final OpenAiChatCompletionOutput result =
     OpenAiClient.forModel(new OpenAiModel("model")).chatCompletion(request);
 ```
 
+### Stream chat completion
+
```
MatKuhr, line 208: (minor) I think it would be helpful to describe in 2 sentences what we are documenting here: passing the stream to a frontend.

Maybe it would even be worthwhile to first showcase using the stream on it's own, e.g. simply printing to stdout, and only in a second step show how one can return it to a frontend non-blocking. Probably easier to understand that way 😉

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -105,6 +108,41 @@ public OpenAiChatCompletionOutput chatCompletion(
     return execute("/chat/completions", parameters, OpenAiChatCompletionOutput.class);
   }
 
+  /**
+   * Generate a completion for the given prompt.
+   *
+   * @param parameters the prompt, including messages and other parameters.
+   * @return A stream of chat completions deltas
```
MatKuhr, line 115:
```suggestion
   * @return A stream of chat completion delta elements.
```

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -105,6 +108,41 @@ public OpenAiChatCompletionOutput chatCompletion(
...
+   */
+  @Nonnull
+  public Stream<String> simpleStreamChatCompletion(
+      @Nonnull final OpenAiChatCompletionParameters parameters) throws OpenAiClientException {
+    return streamChatCompletion(parameters)
+        .peek(
+            delta -> {
+              final String finishReason = delta.getFinishReason();
+              if (finishReason != null && finishReason.equals("content_filter")) {
+                throw new OpenAiClientException("Content filter filtered the output.");
```
MatKuhr, line 140: (minor) Here it would be nice to know more details, e.g. the threshold that was breached. But that would also apply to non-streaming, so maybe let's do it in a separate ticket/PR 😄

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * Processes a {@link ClassicHttpResponse} and returns some value corresponding to that response.
+   *
+   * @param response The response to process
+   * @return A {@link Stream} of a model class instantiated from the response
+   * @throws OpenAiClientException in case of a problem or the connection was aborted
+   */
+  @Nonnull
+  public Stream<D> handleResponse(@Nonnull final ClassicHttpResponse response)
```
MatKuhr, line 34:
```suggestion
  Stream<D> handleResponse(@Nonnull final ClassicHttpResponse response)
```

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Nonnull private final Class<D> deltaType;
+
+  /**
+   * Processes a {@link ClassicHttpResponse} and returns some value corresponding to that response.
+   *
+   * @param response The response to process
+   * @return A {@link Stream} of a model class instantiated from the response
+   * @throws OpenAiClientException in case of a problem or the connection was aborted
+   */
```
MatKuhr, line 32:
```suggestion
```

Doesn't look like actually useful javadoc

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
```diff
@@ -0,0 +1,98 @@
...
+      buildExceptionAndThrow(response);
+    }
+    return parseResponse(response);
+  }
+
+  /**
+   * @param response The response to process
+   * @return A {@link Stream} of a model class instantiated from the response
+   * @author stippi
+   */
```
MatKuhr, line 46:
```suggestion
```

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
```diff
@@ -0,0 +1,98 @@
...
+            responseLine -> {
+              if (!responseLine.startsWith("data: ")) {
+                parseErrorAndThrow(
+                    responseLine,
+                    new OpenAiClientException("Failed to parse response from OpenAI model"));
+              }
+            })
+        .map(
+            responseLine -> {
+              final String data = responseLine.substring(5).replace("delta", "message");
```
MatKuhr, line 79: what does the `replace` do?

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
```diff
@@ -0,0 +1,98 @@
...
+              }
+            })
+        .onClose(
+            () -> {
+              try {
+                inputStream.close();
+              } catch (IOException e) {
+                log.error("Could not close HTTP input stream", e);
+              }
+            });
```
MatKuhr, line 96: I'd be curious if this code is actually needed, or if the stream is also closed without it.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -105,6 +108,41 @@ public OpenAiChatCompletionOutput chatCompletion(
...
+   */
+  @Nonnull
+  public Stream<OpenAiChatCompletionDelta> streamChatCompletion(
+      @Nonnull final OpenAiChatCompletionParameters parameters) throws OpenAiClientException {
+    parameters.setStream(true);
+    return streamChatCompletion("/chat/completions", parameters, OpenAiChatCompletionDelta.class);
+  }
+
+  /**
+   * Generate a completion for the given prompt.
```
MatKuhr, line 126: (minor) The javadoc should explain a bit how this is different form `streamChatCompletion`. Also, this simpler variant currently this is not mentioned in the docs.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -105,6 +108,41 @@ public OpenAiChatCompletionOutput chatCompletion(
...
+   *
+   * @param parameters the prompt, including messages and other parameters.
+   * @return A stream of message deltas
+   * @throws OpenAiClientException if the request fails or if the finish reason is content_filter
+   */
+  @Nonnull
+  public Stream<String> simpleStreamChatCompletion(
```
MatKuhr, line 133: How about calling this:

```suggestion
  public Stream<String> streamChatCompletion(
```

and the other one `streamChatCompletionDeltas`?

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
```diff
@@ -3,26 +3,47 @@
...
 import org.assertj.core.api.SoftAssertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.TestInstance.Lifecycle;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.Mockito;
 
+@TestInstance(Lifecycle.PER_CLASS)
```
MatKuhr, line 46: what is the motivation for this change?

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
```diff
@@ -128,122 +159,309 @@ void testErrorHandling() {
...
-
-    assertThat(result.getChoices()).hasSize(1);
-    OpenAiChatCompletionChoice choice = result.getChoices().get(0);
-    assertThat(choice.getFinishReason()).isEqualTo("stop");
-    assertThat(choice.getIndex()).isEqualTo(0);
-    assertThat(choice.getMessage().getContent())
-        .isEqualTo(
-            """
+  void chatCompletion() throws IOException {
+    try (var inputStream =
```
MatKuhr, line 163: why change this? with the additional indentation from the try it looks a bit less pretty than before arguably

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
```diff
@@ -128,122 +159,309 @@ void testErrorHandling() {
...
+                  List.of(
+                      new OpenAiChatUserMessage()
+                          .addText(
+                              "Can you give me the first 100 numbers of the Fibonacci sequence?")));
+
+      try (Stream<OpenAiChatCompletionDelta> stream = client.streamChatCompletion(request)) {
+        assertThatThrownBy(() -> stream.forEach(System.out::println))
+            .isInstanceOf(OpenAiClientException.class)
+            .hasMessage(
+                "Failed to parse response from OpenAI model and error message: 'exceeded token rate limit'");
```
MatKuhr, line 324: do you know if this can actually happen mid-stream?

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
CharlesDuboisSAP: This is actually needed, I tested without it and the test fails
_Note: the stream is only closed by using try-with-resources. consuming the stream fully (terminal operation) does not close it._

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
CharlesDuboisSAP: I am legally required to mention Stephan as the author 🫡

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
```diff
@@ -128,122 +159,309 @@ void testErrorHandling() {
...
+                  List.of(
+                      new OpenAiChatUserMessage()
+                          .addText(
+                              "Can you give me the first 100 numbers of the Fibonacci sequence?")));
+
+      try (Stream<OpenAiChatCompletionDelta> stream = client.streamChatCompletion(request)) {
+
+        OpenAiChatCompletionOutput totalOutput = new OpenAiChatCompletionOutput();
+        final List<OpenAiChatCompletionDelta> deltaList =
+            stream.peek(totalOutput::addDelta).toList();
```
MatKuhr, line 362: To assert the stream isn't consumed prematurely I'd try something like:

```suggestion
final List<OpenAiChatCompletionDelta> deltaList =
            stream.peek(totalOutput::addDelta)
                    .peek(any -> {
                      try {
                        assertThat(inputStream.available() > 0);
                      } catch (IOException e) {
                        throw new RuntimeException(e);
                      }
                    })
                    .toList();

        assertThat(inputStream.available()).isZero();
```

The IO exception is annoying, maybe moving the assertion to a small helper method so that it looks better.

e2e-test-app/src/test/java/com/sap/ai/sdk/app/controllers/OpenAiTest.java
```diff
@@ -23,12 +32,44 @@ void chatCompletionImage() {
     assertThat(message.getContent()).isNotEmpty();
   }
 
+  @Test
+  void streamChatCompletion() {
+    final var request =
+        new OpenAiChatCompletionParameters()
+            .setMessages(List.of(new OpenAiChatUserMessage().addText("Who is the prettiest?")));
```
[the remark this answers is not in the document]
CharlesDuboisSAP: This is not in the azure spec

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
CharlesDuboisSAP: Removed

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
CharlesDuboisSAP: It removes a warning from my IDE. I read all warnings before committing.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
CharlesDuboisSAP: I experienced it once, I wrote it down [here](https://github.com/SAP/ai-sdk-java/pull/25/commits/69ae7eb220f19dd973330d3b6ce0e47c569fd3b7#diff-8f8ce52bc428755b7c06bb07061be2021bf64675307b39101ba85b3549f35019R44-R50)

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
CharlesDuboisSAP: nvm I removed it by making the client static

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
MatKuhr: maybe you can disable the warning for tests? Because here on error the test fails anyway, so closing the stream isn't really necessary..

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
MatKuhr: > consuming the stream fully (terminal operation) does not close it.

Hmm, but if that is the case, how does adding an implementation for `.onClose` change anything?

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
MatKuhr: Why? None of the other authors of this codebase are mentioned.. Not that I have anything against it, it just looks a bit odd IMO 😉

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
MatKuhr: Thanks, I'll attach this to a related BLI I am creating.

e2e-test-app/src/test/java/com/sap/ai/sdk/app/controllers/OpenAiTest.java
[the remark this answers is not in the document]
MatKuhr: But as far as I can tell from my testing: it works anyway 😉

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
CharlesDuboisSAP: Stephan Assmus:
> If possible, please attribute the code to me somehow. :smile:

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
CharlesDuboisSAP: Because try-with-resources closes the stream

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
CharlesDuboisSAP: Mentioned in the docs

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
MatKuhr: ah okay, and the input stream is not closed without this? I would have thought that maybe whatever implementation of stream is the default used under the hood here would "pass on" the close to any underlying streams the final stream is built upon

But if we have to manually close it, why not close the buffered reader instead? [It will close the input stream internally](https://stackoverflow.com/questions/1388602/do-i-need-to-close-both-filereader-and-bufferedreader).

e2e-test-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java
```diff
@@ -38,6 +48,101 @@ public static OpenAiChatCompletionOutput chatCompletion() {
...
+                List.of(
+                    new OpenAiChatUserMessage()
+                        .addText(
+                            "Can you give me the first 100 numbers of the Fibonacci sequence?")));
+
+    final var stream = OpenAiClient.forModel(GPT_35_TURBO).streamChatCompletionDeltas(request);
+
+    final var emitter = new ResponseBodyEmitter();
+
+    final Runnable consumeStream =
```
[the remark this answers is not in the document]
CharlesDuboisSAP: @MatKuhr

e2e-test-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java
```diff
@@ -38,6 +48,101 @@ public static OpenAiChatCompletionOutput chatCompletion() {
...
+        () -> {
+          final var totalOutput = new OpenAiChatCompletionOutput();
+          // try-with-resources ensures the stream is closed
+          try (stream) {
+            stream
+                .peek(totalOutput::addDelta)
+                .forEach(delta -> send(emitter, delta.getDeltaContent()));
+          } finally {
+            send(emitter, "\n\n-----Total Output-----\n\n" + objectToJson(totalOutput));
```
newtork, line 81: (Comment/Minor)

It could be worthwhile to not just dump the object-to-string. But instead to extract the relevant information and write it piece-by-piece. That would also demonstrate the actual benefit of the feature. Right now the interested reader would wunder.. _"what does total output even mean?"_

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiResponseHandler.java
```diff
@@ -85,19 +94,29 @@ private void buildExceptionAndThrow(@Nonnull final ClassicHttpResponse response)
...
     if (error == null) {
-      throw exception;
+      throw baseException;
     }
-    final var message =
-        "Request to OpenAI model failed with %s %s and error message: '%s'"
-            .formatted(response.getCode(), response.getReasonPhrase(), error.getMessage());
-    throw new OpenAiClientException(message);
+    throw new OpenAiClientException(
+        baseException.getMessage() + " and error message: '%s'".formatted(error.getMessage()));
```
newtork, line 120: (Minor)

Mixed string concatenation + format string.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientException.java
CharlesDuboisSAP: I removed it

e2e-test-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java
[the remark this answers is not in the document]
MatKuhr: better readability, if we inline this it's arguably a bit harder to read where the lambda ends and which code exactly is run async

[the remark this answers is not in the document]
newtork: To me the Runnable reference indicated a later use of the reference. But no strong opinion, we can leave it as is.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
```diff
@@ -128,122 +165,309 @@ void testErrorHandling() {
...
-    assertThat(choice.getFinishReason()).isEqualTo("stop");
-    assertThat(choice.getIndex()).isEqualTo(0);
-    assertThat(choice.getMessage().getContent())
-        .isEqualTo(
-            """
+  void chatCompletion() throws IOException {
+    try (var inputStream =
+        getClass().getClassLoader().getResourceAsStream("chatCompletionResponse.json")) {
+
+      assert inputStream != null;
```
newtork, line 172: (Major/Question)

AFAIK usage of "assert" is not accepted.

newtork (approved):

PR #44, opened 2024-09-04
Update SAP Cloud SDK dependency [`5.12.0`]

newtork:
Update Cloud SDK


MatKuhr (approved):

MatKuhr:
We could (should?) also make use of the new unknown enum feature, but we can also add it later in a separate PR ofc.

PR #49, opened 2024-09-04
[Streaming] Fix issues when not using `try-with-resource`

newtork:
Make the streaming API lenient towards `try-with-resources`.

**Before:**
* **with** try-with-resource: HTTP connections are closed in normal cases, response errors, http errors
* **without** try-with-resource: HTTP connections are NOT closed

**After:**
* **with** try-with-resource: HTTP connections are closed in normal cases, response errors, http errors
* **without** try-with-resource: HTTP connections are closed in normal cases, and http errors. Errors in response payload does not lead to closing of HTTP connection.

```diff
- InputStream <> BufferedReader <> Stream
+ InputStream <> BufferedReader <> [HandledIterator] <> Stream
```

The suggested `HandledIterator` understands when `BufferedReader#readLine()` returns `null`, then the underlying `InputStream` shall be closed [left]. Also when the consuming `Stream` is closed or finished iteration [right].

<details><summary>See JavaDoc of <code>String BufferedReader#readLine()</code> to understand <code>null</code> contract.</summary>

image

</details>

<details><summary>See JavaDoc and code of <code>Stream<String> BufferedReader#lines()</code> for comparison.</summary>

image

</details>


CharlesDuboisSAP (changes requested):
You can remove try-with-resources in the controller and readme since it doesn't do anything anymore

e2e-test-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java
```diff
@@ -112,8 +112,11 @@ public static ResponseEntity<ResponseBodyEmitter> streamChatCompletion() {
 
     final Runnable consumeStream =
         () -> {
-          stream.forEach(deltaMessage -> send(emitter, deltaMessage));
-          emitter.complete();
+          try (stream) {
```
newtork, line 115: The `Stream<String>` may `throwOnContentFilter`, without recovery. Therefore it is recommended to keep the try-with-resources.

This does not apply for `Stream<OpenAiChatCompletionDelta>` variant.

e2e-test-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java
```diff
@@ -69,15 +69,10 @@ public static ResponseEntity<ResponseBodyEmitter> streamChatCompletionDeltas() {
...
-          // try-with-resources ensures the stream is closed
-          try (stream) {
-            stream
-                .peek(totalOutput::addDelta)
-                .forEach(delta -> send(emitter, delta.getDeltaContent()));
-          } finally {
-            send(emitter, "\n\n-----Total Output-----\n\n" + objectToJson(totalOutput));
-            emitter.complete();
-          }
+          stream
```
MatKuhr, line 72: as discussed in slack, let's keep the try with resources because emitter might throw

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/StreamConverter.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.hc.core5.http.HttpEntity;
+
+@Slf4j
+class StreamConverter {
+
+  /** see {@link BufferedReader#DEFAULT_CHAR_BUFFER_SIZE} */
+  static final int BUFFER_SIZE = 8192;
+
+  @FunctionalInterface
+  private interface ReadHandler<T> {
```
MatKuhr, line 31: do we really need this? Maybe we can use `Consumer<T>` instead?

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/StreamConverter.java
```diff
@@ -0,0 +1,108 @@
...
+      if (next == null && !hasNext()) {
+        throw new NoSuchElementException();
+      }
+      return next;
+    }
+  }
+
+  @SuppressWarnings("PMD.CloseResource")
+  @Nonnull
+  static Stream<String> streamLines(@Nullable final HttpEntity entity)
```
MatKuhr, line 85: (minor) I was a bit surprised that `streamLines` still comes here after all the inner class code, maybe move this up?

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/StreamConverter.java
```diff
@@ -0,0 +1,108 @@
...
+
+  @RequiredArgsConstructor
+  private static class HandledIterator<T> implements Iterator<T> {
+    private final ReadHandler<T> readHandler;
+    private final CloseHandler stopHandler;
+    private boolean done = false;
+    private T next = null;
+
+    @Override
+    public boolean hasNext() {
```
MatKuhr, line 56: A note somewhere here that states parallel execution is NOT supported (because we don't see a usecase for it) might be useful.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/StreamConverter.java
```diff
@@ -0,0 +1,108 @@
...
+      if (done) {
+        return false;
+      }
+      try {
+        next = readHandler.readEntry();
```
MatKuhr, line 61: I think this is an issue because it makes calling `hasNext()` not idempotent. Calling it multiple times will empty the reader.

With the current implementation it might not be a visible issue, because we control the stream this is wrapped into, and depending how the stream uses `hasNext()` internally it could be fine, but that seems risky.

MatKuhr (commented):
Left some comments. I would consider un-nesting the `HandledIterator` and merge it with the `StreamConverter` into just one class.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/StreamConverter.java
MatKuhr: I tested what happens if one runs: 

```
var i = stream.iterator();
assertThat(i.next()).isEqualTo("THIS");
assertThat(i.hasNext()).isTrue();
assertThat(i.hasNext()).isTrue();
assertThat(i.next()).isEqualTo("IS");
```

on `"THIS\nIS\nA\nTEST\n"` and interestingly it works. It looks like the implementation of `iterator` internally buffers the result of querying the underlying iterator..

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/StreamConverter.java
newtork: I was assuming nested class/type definitions in productive code must be at top of class. But I'm fine with reordering, according to visibility.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/StreamConverter.java
newtork: Consumer does not allow for checked exception.

I considered alternatives, e.g. VAVR exception types. But I appreciated the narrow declaration of thrown exceptions, i.e. only `IOException` in our case.

However, this means if `#readEntry` throws any unchecked exception the stream will not be closed via close-handler. 

Not a strong opinion. I may have to reconsider.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/StreamConverter.java
newtork: I agree. Even though we control the stream/iterator usage, the `Iterator` implementation for multiple `#hasNext()` is flawed technically.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/StreamConverter.java
MatKuhr: While it's not typical, we have no control or guarantee that parallel execution won't happen. But since there is no use case for it we should make the decision to not be thread safe explicit 😉

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java
```diff
@@ -64,15 +64,10 @@ public static ResponseEntity<ResponseBodyEmitter> streamChatCompletionDeltas() {
...
         () -> {
           final var totalOutput = new OpenAiChatCompletionOutput();
-          // try-with-resources ensures the stream is closed
-          try (stream) {
-            stream
-                .peek(totalOutput::addDelta)
-                .forEach(delta -> send(emitter, delta.getDeltaContent()));
-          } finally {
-            send(emitter, "\n\n-----Total Output-----\n\n" + objectToJson(totalOutput));
-            emitter.complete();
```
CharlesDuboisSAP, line 74: Removing `emitter.complete();` breaks the endpoint. Also shouldn't the customer catch errors anyway?

newtork: > Removing `emitter.complete();` breaks the endpoint. Also shouldn't the customer catch errors anyway?

Thanks, this change was not deliberate.

> Also shouldn't the customer catch errors anyway?

Reason for the change is to demonstrate the usage without `try`.
This is for demo / power-point purpose.
Even if we consider this not "best" practice.

So yes, in a perfect world customers catch errors. In a realistic world people forgot / don't care about try-catch on this level.

CharlesDuboisSAP: Then don't forget to also update the other `streamChatCompletions` endpoint

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/StreamConverter.java
newtork: I changed the code in case of repeated `hasNext()`

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/StreamConverter.java
newtork: Added JavaDoc to emphasize on the fact 👍

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/StreamConverter.java
newtork: I changed the code to avoid unnecessary nested types

e2e-test-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java
newtork: My intention was to offer demo / powerpoint-ready code as part of the test app, but I'm also fine with reverting - and sticking to technically-best-practice with try-with-resources.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java
newtork: It's only now that I understood [Matthias concern above](https://github.com/SAP/ai-sdk-java/pull/49#discussion_r1754375596). Since I don't want to have the discussion, I will rather revert my changes :)

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/IterableStreamConverterTest.java
```diff
@@ -0,0 +1,106 @@
...
+              byte[] bytes = TEMPLATE.getBytes(StandardCharsets.UTF_8);
+              for (int i = 0; i < ar.length; i++) ar[i] = bytes[i % bytes.length];
+              return ar.length;
+            })
+        .thenThrow(new IOException("Ups!"));
+
+    final var entity = new InputStreamEntity(inputStream, ContentType.TEXT_PLAIN);
+
+    final var sut = IterableStreamConverter.lines(entity);
+    assertThatCode(sut::count)
```
CharlesDuboisSAP, line 97:
```suggestion
    assertThatThrownBy(sut::count)
```

The main difference with assertThatCode(ThrowingCallable) is that this method fails if no exception was thrown.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/IterableStreamConverterTest.java
```diff
@@ -0,0 +1,106 @@
+package com.sap.ai.sdk.foundationmodels.openai;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
```
CharlesDuboisSAP, line 4:
```suggestion
import static org.assertj.core.api.Assertions.assertThatThrownBy;
```

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiStreamingHandler.java
```diff
@@ -42,20 +36,8 @@ Stream<D> handleResponse(@Nonnull final ClassicHttpResponse response)
   @SuppressWarnings("PMD.CloseResource")
   private Stream<D> parseResponse(@Nonnull final ClassicHttpResponse response)
```
CharlesDuboisSAP, line 37: > The stream is closed by the user of the Stream

This is false now. Could you update this comment and duplicate it in `IterableStreamConverter.lines()`
Alternatively just merge `parseResponse` inside of `handleResponse` since `parseResponse` only contains a `return` and nothing else

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/IterableStreamConverter.java
```diff
@@ -0,0 +1,113 @@
...
+      next = readHandler.call();
+      isNextFetched = true;
+      if (next == null) {
+        isDone = true;
+        stopHandler.run();
+      }
+    } catch (final Exception e) {
+      isDone = true;
+      stopHandler.run();
+      throw new IllegalStateException("Iterator stopped unexpectedly.", e);
```
CharlesDuboisSAP, line 68: What can be thrown here?

CharlesDuboisSAP (commented):
I get a timeout on `http://localhost:8080/streamChatCompletionDeltas`

MatKuhr:
Let's finish this PR up so that we are free to continue with streaming for orchestration

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/IterableStreamConverter.java
newtork: The exception caught here would be thrown by `readHandler.call()`.
In this case that would be from `BufferedReader#readLine`.
The `readLine` may throw `IOException – If an I/ O error occurs`.
Presumably this would happen when the nested `InputStream#read` would fail.

One possibility for an `InputStream` from `HttpEntity` to fail would be a **network interruption**.

newtork:
> I get a timeout on `http://localhost:8080/streamChatCompletionDeltas`

I was not able to reproduce the timeout. Please kindly recheck.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/IterableStreamConverter.java
CharlesDuboisSAP: Then either do not throw anything or update the expection catching IOException and with a message like `Stream closed`. This is not an `IllegalStateException` if it can happen

newtork: I agree, network issues will occur someday anyway. So let's improve error messages. https://github.com/SAP/ai-sdk-java/pull/49/commits/fb21544f9c622a38d08765066e545a585a34ae64

CharlesDuboisSAP (approved):

PR #64, opened 2024-09-19
Added release script

CharlesDuboisSAP:
## Context

AI/ai-sdk-java-backlog#65.

Have a release script ready and tested before our release date.

### Feature scope:
 
- [x] Can be compared to the Cloud SDK release script
  - [Prepare release successful run](https://github.com/CharlesDuboisSAP/ai-sdk-java/actions/runs/11066130980)
  - [Perform release successful run](https://github.com/CharlesDuboisSAP/ai-sdk-java/actions/runs/11066498951) (apart for the nexus deploy failing on package names)
  - [Release PR created](https://github.com/CharlesDuboisSAP/ai-sdk-java/pull/6)

## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~


pom.xml
```diff
@@ -356,7 +381,7 @@
             <goals>
               <goal>apply</goal>
             </goals>
-            <phase>compile</phase>
+            <phase>generate-sources</phase>
```
newtork, line 384: (Minor/Question)

Shouldn't spotless run on "process-sources", i.e. after "generate-sources"?

sample-code/spring-app/pom.xml
```diff
@@ -189,6 +188,30 @@
           <skipTests>${skipTests}</skipTests>
         </configuration>
       </plugin>
+      <!-- Skip release -->
```
newtork, line 191: (Minor)

I don't really like how complex it is for disabling a module for release. Can you instead reduce it to just a _property_ like `<release.skip>true</release.skip>` in this file and default `false` in parent file(?)

If yes, then you would add the `<skip>${release.skip}</skip>` to parent pom.

pom.xml
CharlesDuboisSAP: You're right, corrected

sample-code/spring-app/pom.xml
```diff
@@ -189,6 +188,30 @@
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <executions>
+          <execution>
+            <goals>
+              <goal>deploy-file</goal>
```
newtork, line 198: (Minor/Question)

If I understood correctly you are skipping both for `<goals><goal>deploy-file` and `<goals/>`. Isn't the former redundant? I'm not sure when it comes to Maven plugin configurations.

sample-code/spring-app/pom.xml
CharlesDuboisSAP: That would be great if the `maven-deploy-plugin` had a `skip` property...
Not all plugins have skip properties.

sample-code/spring-app/pom.xml
CharlesDuboisSAP: During my testing I found that both needed to be present to correctly skip.

.pipeline/scripts/set-release-versions.py
```diff
@@ -0,0 +1,89 @@
...
+        self.end(ElementTree.Comment)
+
+def _update_readme(sdk_version):
+    _update_file("README.md", r'(maven_central-)(.+)(-blue\.svg)', r'\g<1>%s\g<3>' % sdk_version)
+    _update_file("README.md", r'(a:sdk-core%20AND%20v:)(.+\))', r'\g<1>%s)' % sdk_version)
+
+def _update_pom_files(sdk_version):
+    _update_version_tags(sdk_version)
+
+def _update_version_tags(sdk_version):
```
newtork, line 23: (Minor)

Not sure why delegating to another method. What's the point of having `update_pom_files`?

.pipeline/scripts/set-release-versions.py
```diff
@@ -0,0 +1,89 @@
...
+        super(_CommentedTreeBuilder, self).__init__(*args, **kwargs)
+
+    def comment(self, data):
+        self.start(ElementTree.Comment, {})
+        self.data(data)
+        self.end(ElementTree.Comment)
+
+def _update_readme(sdk_version):
+    _update_file("README.md", r'(maven_central-)(.+)(-blue\.svg)', r'\g<1>%s\g<3>' % sdk_version)
+    _update_file("README.md", r'(a:sdk-core%20AND%20v:)(.+\))', r'\g<1>%s)' % sdk_version)
```
newtork, line 18: (Comment)

This looks like it should better be a regular grep/sed command than a python script :)

.pipeline/scripts/get-release-versions.py
```diff
@@ -0,0 +1,52 @@
+import os
```
newtork, line 1: (Comment)

I'm sorry for all the python fans, but this could've been a single CLI line, like:
```
mvn help:evaluate -Dexpression=project.version -q -DforceStdout
```

But I'm okay with leaving the XML parsing, since this is what we do in the `set-release-version.py` anyway.

.pipeline/scripts/set-release-versions.py
```diff
@@ -0,0 +1,89 @@
...
+    if len(project_version_tags) == 1:
+        project_version_tags[0].text = sdk_version
+
+    parent_tags = root.findall(".//xmlns:parent", namespaces=namespaces)
+
+    if len(parent_tags) == 1:
+        parent_tags[0].find("xmlns:version", namespaces=namespaces).text = sdk_version
+
+    # not sure why, but tree.write() (as used in other functions) produces a differently formatted XML declaration...
+    # that's why we the file is written in this kind of hacky manner
```
newtork, line 60:
```suggestion
    # that's why the file is written in this kind of hacky manner
```

.pipeline/scripts/set-release-versions.py
CharlesDuboisSAP: This is straight from the Cloud SDK, should I remove it?

newtork: Nevermind, I realize this is a recursive function. Hm..

.pipeline/scripts/set-release-versions.py
newtork: Nevermind, you're just reusing the existing method.

Personally I don't see the benefit of using Python over `grep` / `sed`. But that's not a hill I'm planning to die on :)

.github/actions/pr-is-mergeable/action.yaml
```diff
@@ -0,0 +1,122 @@
...
+    required: false
+    default: ${{ github.repository }}
+  token:
+    description: "The GitHub access token (with PR read permissions) to access the PR"
+    required: false
+    default: ${{ github.token }}
+  excluded-check-runs:
+    description: "A JSON object, where the keys are the names of the workflows to exclude. The values are a list of step names to ignore. If the list is empty, all steps of the workflow will be ignored."
+    required: false
+    default: '{\"${{ github.workflow }}\": []}'
```
newtork, line 23: (Question)

Doesn't this mean by default "all steps of workflow ${{ github.workflow }}" will be ignored? This doesn't have an effect on the action? Wouldn't default "{}" better fit, i.e. to not ignore anything?

CharlesDuboisSAP: This whole file is unreadable. If it was approved for the Cloud SDK we can also approve it of the AI SDK.

.github/workflows/perform-release.yml
```diff
@@ -0,0 +1,142 @@
...
+        uses: actions/checkout@v4
+
+      - name: "Check Whether Code PR Can Be Merged"
+        if: ${{ inputs.skip-pr-merge != 'true' }}
+        uses: ./.github/actions/pr-is-mergeable
+        with:
+          pr-ref: ${{ steps.determine-branch-names.outputs.CODE_BRANCH_NAME }}
+          excluded-check-runs: |
+            {
+              \"Continuous Integration\": [\"Run BlackDuck Scan\", \"Run Security Rating\"],
```
newtork, line 67: (Minor/Question)

This is a leftover from copying right? We don't use Blackduck in this repo, AFAIK.

.github/workflows/perform-release.yml
```diff
@@ -0,0 +1,142 @@
...
+        env:
+          MAVEN_CENTRAL_USER: ${{ secrets.MAVEN_CENTRAL_USER }}
+          MAVEN_GPG_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
+
+      - name: "Deploy Staging"
+        run: |
+          # Create an empty directory to correctly execute the plugin
+          mkdir temp
+          cd temp
+          MVN_ARGS="$MVN_CLI_ARGS org.sonatype.plugins:nexus-staging-maven-plugin:1.7.0:deploy-staged-repository -DserverId=ossrh -DnexusUrl=https://oss.sonatype.org -DrepositoryDirectory=../temp_local_repo -DstagingProfileId=$MAVEN_CENTRAL_PROFILE_ID"
```
newtork, line 126: (Question)

`-DnexusUrl` is required? What does it do?

.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,196 @@
...
+      - name: "Setup java"
+        uses: actions/setup-java@v4
+        with:
+          distribution: "temurin"
+          java-version: ${{ env.JAVA_VERSION }}
+          cache: 'maven'
+
+      - name: "Build SDK"
+        run: |
+          MVN_ARGS="${{ env.MVN_MULTI_THREADED_ARGS }} clean install -DskipTests -DskipFormatting"
```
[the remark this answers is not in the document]
CharlesDuboisSAP: No, I removed it.

.github/workflows/perform-release.yml
CharlesDuboisSAP: Yes, you can release to different repositories

.github/actions/pr-is-mergeable/action.yaml
newtork: I thought you had higher standards for AI SDK 😁

CharlesDuboisSAP: I removed the whole generation of pom files with plugins script. Maybe we get rid of this in 2030 😁

.github/workflows/perform-release.yml
newtork: It confused me that for server selection we actively use reserved argument `-DserverId` (resulting in `https://oss.sonatype.org` implicitly). While additionally we have custom(?) argument `-DnexusUrl` set to `https://oss.sonatype.org` explicitly. This looks like duplicate information.

But since we have custom plugin `nexus-staging-maven-plugin` executing `deploy-staged-repository`, I have to assume this argument is originating from that third party.

newtork (commented):
Apart from the deploy-exclusion logic, I would approve the PR.

Maybe we can explore other options than explicitly "skipping" two specific executions of one specific plugin. Like [here](https://github.com/SAP/ai-sdk-java/pull/77/files).

newtork (approved):

PR #77, opened 2024-09-30
[Re: Added release script] Potential simplification

newtork:
I think it would be safer to exclude the whole sub-module when deploying.

For that we would change the cli usage from `P` profile to `-D` system arg.


CharlesDuboisSAP (commented):
This is cleaner, I like it. But I can't test the `-Drelease` option since the `gpg-plugin` is now broken for me

PR #53, opened 2024-09-06
Added deployment cache

CharlesDuboisSAP:
## Context

AI/ai-sdk-java-backlog#54.

The lookup of deployments should be more efficient & resilient.
#### Note for reviewers:
The filter for Orchestration and OpenAI deployments has been removed since the new `AiCoreService` allows to fetch all deployments

### Feature scope:
 
- [x] Cache for deployments
  - Initialized lazily or eagerly whenever the customer calls `new AiCoreService().forDeployment...`
  - No timeout
  - Customers can call `DeploymentCache.resetCache()` whenever they delete a deployment

## Definition of Done

- [x] Functionality scope stated & covered
- [x] Tests cover the scope above
- [x] Error handling created / updated & covered by the tests above
- [x] Aligned changes with the JavaScript SDK
- [ ] ~Documentation updated~
- [ ] ~Release notes updated~

PR #79, opened 2024-10-01
Refactor `Core` class

newtork:
BLI
https://github.tools.sap/AI/ai-sdk-java-backlog/issues/63

Core motivations:
* `Core` to `AiCoreService` class
* Expose `client()` and `detination()`
* (Allow for **LAZY EVALUATION**)
* Allow for custom destinations
* Allow for convenient deployment selection
* Allow for overriding and customization to enable future workarounds
* Make resource group optional

```java
private static final OrchestrationCompletionApi API =
    new OrchestrationCompletionApi(
        new AiCoreService()
            .forDeploymentByScenario("orchestration")
            .withResourceGroup("default") // optional
            .client());
```

```java
private static final ScenarioApi API = new ScenarioApi(new AiCoreService().client());
```

Auto completion:

<details><summary>new AiCoreService()</summary>

image

</details>
<details><summary>.forDeploymentByScenario("orchestration")</summary>

image

</details>
<details><summary>.withResourceGroup("foo")</summary>

image

</details>
<details><summary>.client()</summary>

```ApiClient```

</details>

## Definition of Done

- [x] ~Documentation updated~
- [x] ~Release notes updated~

### CODE REVIEW

File: core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java [hunk 1]
```diff
@@ -0,0 +1,95 @@
...
+import java.util.function.Supplier;
+import javax.annotation.Nonnull;
+import lombok.AccessLevel;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+/** Connectivity convenience methods for AI Core. */
+@Slf4j
+@RequiredArgsConstructor(access = AccessLevel.PROTECTED)
+public class AiCoreService implements AiCoreServiceStub {
```
CharlesDuboisSAP, line 20: What made you change from `AiClient` _(or `AiCoreClient`)_ to `AiCoreService`.
The service is in BTP, we only offer a client.

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /**
+   * Set a specific base destination.
+   *
+   * @param destination The destination to be used for AI Core service calls.
+   * @return A new instance of the AI Core service.
+   */
+  @Nonnull
+  public AiCoreService withDestination(@Nonnull final Destination destination) {
```
CharlesDuboisSAP, line 43: I would rather have:
```diff
-withDestination()
+forDestination()
```
```diff
-withDeploymentByModel()
+forDeploymentWithModel()   or forModel()...
```
```diff
-withDeploymentByScenario()
+forDeploymentWithScenario()
```

core/src/main/java/com/sap/ai/sdk/core/AiCoreServiceStub.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /** Options for the API client. */
+  interface ClientOptions {
+
+    /** Serialize with null values. */
+    ClientOptions SERIALIZE_WITH_NULL_VALUES = () -> ApiClient::new;
+
+    /** Serialize without null values. */
+    ClientOptions SERIALIZE_WITHOUT_NULL_VALUES = () -> ClientOptions::withoutNull;
```
CharlesDuboisSAP, line 59: Why? Do you see any other usecase outside of those 2?

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/DeploymentController.java
```diff
@@ -25,7 +24,7 @@
 @RequestMapping("/deployments")
 class DeploymentController {
 
-  private static final DeploymentApi API = new DeploymentApi(getClient());
+  private static final DeploymentApi API = new DeploymentApi(new AiCoreService().client());
```
CharlesDuboisSAP, line 27:
```suggestion
  private static final DeploymentApi API = new DeploymentApi(new AiCoreService().getClient());
```

How about classic getters for `client()` and `destination()`. Especially if `Client` is not part of the class name of `AiCoreService`

core/src/main/java/com/sap/ai/sdk/core/AiCoreServiceWithDeployment.java
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * Set the resource group.
+   *
+   * @param resourceGroup The resource group handler.
+   * @return A new instance of the AI Core service.
+   */
+  @Nonnull
+  public AiCoreServiceWithDeployment withResourceGroup(
+      @Nonnull final Function<AiCoreServiceWithDeployment, String> resourceGroup) {
+    return new AiCoreServiceWithDeployment(deploymentId, destination, resourceGroup);
```
CharlesDuboisSAP, line 71: This is really confusing!
Is this supposed to be public?
How even do you use it? Do you have a customer side example?

core/src/main/java/com/sap/ai/sdk/core/AiCoreServiceWithDeployment.java
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * Set the destination.
+   *
+   * @param destination The destination handler.
+   * @return A new instance of the AI Core service.
+   */
+  @Nonnull
+  public AiCoreServiceWithDeployment withDestination(
+      @Nonnull final Function<AiCoreServiceWithDeployment, Destination> destination) {
+    return new AiCoreServiceWithDeployment(deploymentId, destination, resourceGroup);
```
CharlesDuboisSAP, line 94: Same, confusing

CharlesDuboisSAP (commented):
From the customer POV, I would only rename the API, it's pretty intuitive in our use-cases.
From a maintainer of the backend POV (me), it's complex and large, I fear maintenance is going to take a long time because of the complexity. I would make the backend simpler and have less API available.

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
newtork: While I understand and mostly agree, the "wither" pattern is [established](https://projectlombok.org/api/lombok/With) in Java world - I think.

Maybe we can avoid mixing "with" and "for" method prefix(?)

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/DeploymentController.java
newtork: * _Power-point readiness_ dictates short API
* Omitting `get` prefix for immutable classes is (modern) common practice, see [Java record classes](https://docs.oracle.com/en/java/javase/17/language/records.html) for example

core/src/main/java/com/sap/ai/sdk/core/AiCoreServiceWithDeployment.java
newtork: * `AiCoreService` offers **lazy** evaluation of `destination()` and `client()`
  * Destinations are not resolved in advance. This avoid unnecessary compuation.
* `AiCoreServiceWithDeployment` is subclass of  `AiCoreService`
  * Due to lazy evaluation, the class stores **no literals** - only handlers.
  * The accessors and constructors allow for setting the handlers.
  * **Convenience methods** allow **for literals**, they are translated to handlers.
  * (Why `Function` instead of `Supplier`? Because some of those literals dynamically depend on destination and deployment.)

We definitely can make these methods `protected` to not interfere with auto-completion!
If users see necessity for workarounds they can still extend the class and access `protected` code.

newtork: > How even do you use it? Do you have a customer side example?

Resolve a resource group id dynamically.

```diff
  private static final OrchestrationCompletionApi API =
      new OrchestrationCompletionApi(
          new AiCoreService()
              .withDeploymentByScenario("orchestration")
-             .withResourceGroup("default")
+             .withResourceGroup(OrchestrationController::getMyResourceGroupId)
              .client());

+ private static String getMyResourceGroupId(AiCoreServiceWithDeployment core) {
+   return new ResourceGroupApi(core.client())
+       .getAll().getResources().stream()
+           .filter(r -> "foobar-tenant".equalsIgnoreCase(r.getTenantId()))
+           .map(BckndResourceGroup::getResourceGroupId)
+           .findAny()
+           .orElseThrow();
+ }
```

newtork: While thinking about it, I think you're right:
* Resource Group Id is not dependent on `deploymentId`.
* Also using the `destination` with "altered" path, any lookup will AI Core interaction will fail, e.g. `ResourceGroupApi`.
* If the user can neither leverage provided destination nor deployment, the handler itself becomes less convenient than just expecting a `String` literal.

newtork: I defused the code.
```diff
  Function<AiCoreServiceWithDeployment, String> deploymentId

- Function<AiCoreServiceWithDeployment, Destination> destination
+ Supplier<Destination> destination

- Function<AiCoreServiceWithDeployment, String> resourceGroup
+ String resourceGroup
```

core/src/main/java/com/sap/ai/sdk/core/AiCoreServiceWithDeployment.java
newtork: I defused the code.
```diff
  Function<AiCoreServiceWithDeployment, String> deploymentId

- Function<AiCoreServiceWithDeployment, Destination> destination
+ Supplier<Destination> destination

- Function<AiCoreServiceWithDeployment, String> resourceGroup
+ String resourceGroup
```

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * Set a specific deployment by id.
+   *
+   * @param deploymentId The deployment id to be used for AI Core service calls.
+   * @return A new instance of the AI Core service.
+   */
+  @Nonnull
+  public AiCoreServiceWithDeployment withDeployment(@Nonnull final String deploymentId) {
+    return new AiCoreServiceWithDeployment(c -> deploymentId, this::destination);
+  }
```
CharlesDuboisSAP, line 56: Any use-case for `withDeployment(deploymentId)`? I would rather less API to support than more.

core/src/main/java/com/sap/ai/sdk/core/AiCoreServiceWithDeployment.java
```diff
@@ -0,0 +1,156 @@
...
+  /**
+   * Create a new instance of the AI Core service with a specific deployment id and destination.
+   *
+   * @param deploymentId The deployment id handler.
+   * @param destination The destination handler.
+   */
+  public AiCoreServiceWithDeployment(
+      @Nonnull final Function<AiCoreServiceWithDeployment, String> deploymentId,
+      @Nonnull final Supplier<Destination> destination) {
+    this(deploymentId, destination, null);
```
CharlesDuboisSAP, line 40:
```suggestion
  @Nonnull private final String resourceGroup;

  /**
   * Create a new instance of the AI Core service with a specific deployment id and destination.
   *
   * @param deploymentId The deployment id handler.
   * @param destination The destination handler.
   */
  public AiCoreServiceWithDeployment(
      @Nonnull final Function<AiCoreServiceWithDeployment, String> deploymentId,
      @Nonnull final Supplier<Destination> destination) {
    this(deploymentId, destination, "default");
```

Nullable should be avoided

CharlesDuboisSAP (commented):
The Javadoc needs to be reworked

core/src/main/java/com/sap/ai/sdk/core/AiCoreServiceStub.java
CharlesDuboisSAP: Everything can be achieved by creating your own implementation of `AiCoreServiceStub` (or extending `AiCoreService`) and changing the implementation of `client()`.
This would cut the `AiCoreServiceStub` in half.

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
MatKuhr: Yes, `AiCoreService` represents and is used for the entire AI Core service, while we have several clients (orchestration, openAi, DeploymentAPI etc.)

core/src/main/java/com/sap/ai/sdk/core/AiCoreServiceStub.java
```diff
@@ -0,0 +1,98 @@
...
+import javax.annotation.Nonnull;
+import org.springframework.http.client.BufferingClientHttpRequestFactory;
+import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
+import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
+import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
+import org.springframework.web.client.RestTemplate;
+
+/** Container for an API client and destination. */
+@FunctionalInterface
+public interface AiCoreServiceStub {
```
MatKuhr, line 21: The naming `Stub` sounds a little bit like test code. I get that we don't want to unnecessarily make customer code more verbose by having the class be named `DefaultAiCoreService`. We could consider `AiCoreServiceI` or maybe simply `AiService`. No strong opinion.

core/src/main/java/com/sap/ai/sdk/core/AiCoreServiceWithDeployment.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import lombok.AccessLevel;
+import lombok.RequiredArgsConstructor;
+
+/** Connectivity convenience methods for AI Core with deployment. */
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+public class AiCoreServiceWithDeployment implements AiCoreServiceStub {
```
MatKuhr, line 20:
```suggestion
public class AiCoreDeployment implements AiCoreServiceStub {
```

For me the `withDeployment` sounds a bit off, because there are many deployments on one AI Core service instance, and the semantic here is actually "withThisSingleDeploymentButWithoutAllTheOtherDeployments" 😄

newtork: Sounds like, similar to [Charles idea](https://github.com/SAP/ai-sdk-java/pull/79#discussion_r1782649790), you would prefer the following method name changes, right?
```diff
- withDeployment[...]
+ forDeployment[...]
```

core/src/main/java/com/sap/ai/sdk/core/DestinationResolver.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.HashMap;
+import java.util.List;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import lombok.extern.slf4j.Slf4j;
+
+/** Utility class to resolve the destination pointing to the AI Core service. */
+@Slf4j
+class DestinationResolver {
+
```
MatKuhr, line 25: How about: 
```suggestion
  private ServiceBindingAccessor accessor = DefaultServiceBindingAccessor.getInstance();
```
for easier unit testing?

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
MatKuhr: I think the "wither" pattern is typically used for changing an existing instance, where `for` is more a static factory prefix. So to me `withDestination()` would be fine, but I'd probably look for something else instead of e.g.  `withDeploymentByScenario()` . See also [this comment](https://github.com/SAP/ai-sdk-java/pull/79/files?diff=unified&w=1#r1782942027).

MatKuhr (commented):
In general I like the approach and I think it already looks significantly improved!

Main thing I would consider is the `AiCoreServiceWithDeployment` class (maybe renamed to `AiCoreDeployment`. Because semantically what it represents is an inference endpoint, which can be e.g. orchestration, an OpenAI model or some custom model.

So we have e.g. 
* `AiCoreService` -> `AiCoreDeployment` -> `OrchestrationClient`
* Credentials, base URL etc. -> path: /inference/id1234/ -> path: /orchestration, body: {...}

I'm not sure if `AiCoreDeployment` should really implement `AiCoreServiceStub`. But I do think it makes sense to have `AiCoreService.getDeployment(String model)`

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
CharlesDuboisSAP: with means in addition. What we are doing is replacing the original URL, and creating a `new AiCoreService()`. which is kind of a static factory.

core/src/main/java/com/sap/ai/sdk/core/DestinationResolver.java
newtork: Yeah sure, we didn't test this in the past and I wasn't planning to add test.

I'm not happy with the class in general, but I didn't want to change it as part of this ticket / PR.

core/src/main/java/com/sap/ai/sdk/core/AiCoreServiceStub.java
newtork: Current suggestion is:
```
AiCoreService extends AiCoreDestination
AiCoreDeployment extends AiCoreDestination
```

> Why not `OrchestrationClient` as suggested [below](https://github.com/SAP/ai-sdk-java/pull/79#pullrequestreview-2340551315)?

Because it would conflict with `OpenAI` client (and future) use cases-

> Why "Destination"?

Because that's the primary value to extract, via `destination()`. Allowing `client()` is only for convenience. Furthermore we may challenge the whole _SAP Cloud SDK_ type `ApiClient` related to that, in mid future - before release maybe.

core/src/main/java/com/sap/ai/sdk/core/AiCoreServiceStub.java
newtork: In general I would agree with you, users providing their implementation of `client()` would be preferred over additional/optional `client(options)`.

Unfortunately, due to the immutable nature of the classes, all accessors, like `withResourceGroup(String)` will instantiate a new object, that escapes the custom/overridden methods.

If the user still wanted to override while having all features, the code would look like the following, given the current state:

<details><summary>(1) base-case, assuming productive code does not make use of deployment API</summary>

```java
class CustomAiCoreServoce extends AiCoreService {
    @Nonnull
    @Override
    public ApiClient client() {
        return new ApiClient(destination()); // re-activates null-serialization
    }
}
```

</details>
<details><summary>(2) extended case, user knows deployment id</summary>

```java
final AiCoreService base = new AiCoreService();
class CustomAiCoreDeployment extends AiCoreDeployment {
    public CustomAiCoreDeployment() {
        super(res -> "my-deployment-id", base::destination);
    }

    @Nonnull
    @Override
    public ApiClient client() {
        return new ApiClient(destination()); // re-activates null-serialization
    }
}
```

</details>
<details><summary>(3) extended case, user knows deployment is identified via scenario</summary>

```java
final AiCoreService base = new AiCoreService();
class OtherAiCoreDeployment extends AiCoreDeployment {
    public OtherAiCoreDeployment() {
        super(res -> getDeploymentId(base.client(), res), base::destination);
    }

    @Nonnull
    @Override
    public ApiClient client() {
        return new ApiClient(destination()); // re-activates null-serialization
    }

    @Nonnull
    static String getDeploymentId(@Nonnull final ApiClient client, @Nonnull final String res) {
        final Predicate<AiDeployment> p = deployment -> "orchestration".equals(deployment.getScenarioId());
        return AiCoreDeployment.getDeploymentId(client, res, p);
    }
}
```

</details>

While the base-case looks fine for me, the extended cases are messy.

Also I'm not sure how much effort we should put further into **the underlying issues** with `ApiClient` from _SAP Cloud SDK_, since we are planning to explore alternatives soon. According to our backlog.

core/src/main/java/com/sap/ai/sdk/core/AiCoreDeployment.java
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  /**
+   * Update and set the default request headers for the destination.
+   *
+   * @param builder The destination builder.
+   * @param dest The original destination reference.
+   */
+  protected void destinationSetHeaders(
+      @Nonnull final DefaultHttpDestination.Builder builder, @Nonnull final Destination dest) {
```
CharlesDuboisSAP, line 131:
```suggestion
   */
  protected void destinationSetHeaders(@Nonnull final DefaultHttpDestination.Builder builder) {
```

Parameter dest unused

core/src/main/java/com/sap/ai/sdk/core/AiCoreDeployment.java
```diff
@@ -0,0 +1,224 @@
...
+    return new AiCoreDeployment(service, obj -> deploymentId);
+  }
+
+  @Nonnull
+  @Override
+  public Destination destination() {
+    final var dest = service.baseDestinationHandler.apply(service);
+    final var builder = service.builderHandler.apply(service, dest);
+    destinationSetUrl(builder, dest);
+    destinationSetHeaders(builder, dest);
```
CharlesDuboisSAP, line 98:
```suggestion
    destinationSetHeaders(builder);
```

core/src/main/java/com/sap/ai/sdk/core/AiCoreDeployment.java
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  /**
+   * Set the destination.
+   *
+   * @param destination The destination.
+   * @return A new instance of the AI Core service.
+   */
+  @Nonnull
+  public AiCoreDeployment withDestination(@Nonnull final Destination destination) {
```
CharlesDuboisSAP, line 153: Do we need this? If yes we should test that this works like expected.
Also I can't understand from the Javadoc, please update .

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
```diff
@@ -0,0 +1,164 @@
...
+    final var destination = destination();
+    return clientHandler.apply(this, destination);
+  }
+
+  @Nonnull
+  @Override
+  public Destination destination() {
+    final var dest = baseDestinationHandler.apply(this);
+    return builderHandler.apply(this, dest).build();
+  }
```
CharlesDuboisSAP, line 58: Can we reduce code duplication between `AiCoreService` and `AiCoreDeployment`? (not by adding methods but changing the structure)

core/src/main/java/com/sap/ai/sdk/core/AiCoreDeployment.java
```diff
@@ -0,0 +1,224 @@
...
+  @Nonnull
+  private final String resourceGroup;
+
+  /**
+   * Default constructor with "default" resource group.
+   *
+   * @param service The AI Core service.
+   * @param deploymentId The deployment id handler.
+   */
+  public AiCoreDeployment(
```
CharlesDuboisSAP, line 45:
```suggestion
  protected AiCoreDeployment(
```

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
```diff
@@ -0,0 +1,164 @@
...
+  public AiCoreService withDestination(@Nonnull final Destination destination) {
+    return new AiCoreService((service) -> destination, clientHandler, builderHandler);
+  }
+
+  /**
+   * Set a specific deployment by id.
+   *
+   * @param deploymentId The deployment id to be used for AI Core service calls.
+   * @return A new instance of the AI Core service.
+   */
```
CharlesDuboisSAP, line 76: Should it be `@return A new instance of the AI Core deployment.` instead?
_Note: same for `forDeploymentByModel`_

core/src/main/java/com/sap/ai/sdk/core/AiCoreDestination.java
```diff
@@ -0,0 +1,28 @@
+package com.sap.ai.sdk.core;
+
+import com.sap.cloud.sdk.cloudplatform.connectivity.Destination;
+import com.sap.cloud.sdk.services.openapi.apiclient.ApiClient;
+import javax.annotation.Nonnull;
+
+/** Container for an API client and destination. */
+@FunctionalInterface
+public interface AiCoreDestination {
```
MatKuhr, line 9: I think we might be able to omit this interface, because we are refactoring orchestration to no longer need an `ApiClient`, and going forward that will probably hold true for any other generated code as well.

Thus, currently the only place we need the customized ApiClient is for the AI API, but for that we already have the convenience method, since its in the same maven module.

So maybe we can simplify. That also saves us from thinking about naming, because for me `AiCoreDestination` is too close to `Destination`, without actually implementing the interface.

core/src/test/java/com/sap/ai/sdk/core/client/WireMockTestServer.java
```diff
@@ -16,12 +18,14 @@ abstract class WireMockTestServer {
 
   static WireMockServer wireMockServer;
   static Destination destination;
+  static ApiClient client;
 
   @BeforeAll
   static void setup() {
     wireMockServer = new WireMockServer(WIREMOCK_CONFIGURATION);
     wireMockServer.start();
     destination = DefaultHttpDestination.builder(wireMockServer.baseUrl()).build();
```
CharlesDuboisSAP, line 27:
```suggestion
  static ApiClient client;

  @BeforeAll
  static void setup() {
    wireMockServer = new WireMockServer(WIREMOCK_CONFIGURATION);
    wireMockServer.start();
    final var destination = DefaultHttpDestination.builder(wireMockServer.baseUrl()).build();
```

core/src/test/java/com/sap/ai/sdk/core/client/WireMockTestServer.java
```diff
@@ -4,8 +4,10 @@
 
 import com.github.tomakehurst.wiremock.WireMockServer;
 import com.github.tomakehurst.wiremock.core.WireMockConfiguration;
+import com.sap.ai.sdk.core.AiCoreService;
 import com.sap.cloud.sdk.cloudplatform.connectivity.DefaultHttpDestination;
 import com.sap.cloud.sdk.cloudplatform.connectivity.Destination;
```
CharlesDuboisSAP, line 9:
```suggestion
```

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -58,7 +58,11 @@ public final class OpenAiClient {
    */
   @Nonnull
   public static OpenAiClient forModel(@Nonnull final OpenAiModel foundationModel) {
-    final var destination = Core.getDestinationForModel(foundationModel.model(), "default");
+    final var destination =
+        new AiCoreService()
+            .forDeploymentByModel(foundationModel.model())
+            .withResourceGroup("default")
+            .destination();
```
MatKuhr, line 65: It might be useful to store the AI Core service itself, not the destination. That way we get the destination on demand, which might be useful if we ever need the destination to change.

orchestration/src/test/java/com/sap/ai/sdk/orchestration/client/OrchestrationUnitTest.java
```diff
@@ -78,15 +81,40 @@ public class OrchestrationUnitTest {
...
   void setup(WireMockRuntimeInfo server) {
+
+    stubFor(
+        get(urlPathEqualTo("/v2/lm/deployments"))
+            .withHeader("AI-Resource-Group", equalTo("default"))
+            .willReturn(
+                aResponse()
+                    .withStatus(HttpStatus.SC_OK)
+                    .withHeader("content-type", "application/json")
+                    .withBody(
```
MatKuhr, line 92:
```suggestion
                    .okJson(
```

core/src/test/java/com/sap/ai/sdk/core/AiCoreServiceTest.java
```diff
@@ -0,0 +1,143 @@
...
+    final var destination = core.destination();
+    final var client = core.client();
+
+    // verification
+    assertThat(destination.get(DestinationProperty.URI)).contains("https://srv/v2/");
+    assertThat(destination.get(DestinationProperty.AUTH_TYPE)).isEmpty();
+    assertThat(destination.get(DestinationProperty.NAME)).singleElement(STRING).contains("aicore");
+    assertThat(destination.get(AI_CLIENT_TYPE_KEY)).contains(AI_CLIENT_TYPE_VALUE);
+    assertThat(client.getBasePath()).isEqualTo("https://srv/v2/");
+    verify(accessor, times(2)).getServiceBindings();
```
CharlesDuboisSAP, line 76: Why 2 times?

orchestration/src/test/java/com/sap/ai/sdk/orchestration/client/OrchestrationUnitTest.java
```diff
@@ -78,15 +81,40 @@ public class OrchestrationUnitTest {
 
   @BeforeEach
   void setup(WireMockRuntimeInfo server) {
+
+    stubFor(
+        get(urlPathEqualTo("/v2/lm/deployments"))
+            .withHeader("AI-Resource-Group", equalTo("default"))
+            .willReturn(
+                aResponse()
```
MatKuhr, line 89:
```suggestion
                ok()
```

Not sure if it makes any difference though

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
```diff
@@ -0,0 +1,164 @@
...
+  public Destination destination() {
+    final var dest = baseDestinationHandler.apply(this);
+    return builderHandler.apply(this, dest).build();
+  }
+
+  /**
+   * Set a specific base destination.
+   *
+   * @param destination The destination to be used for AI Core service calls.
+   * @return A new instance of the AI Core service.
```
CharlesDuboisSAP, line 64:
```suggestion
   * @return A new instance of the AI Core destination.
```
Not sure about this, you can reword it.

core/src/test/java/com/sap/ai/sdk/core/client/WireMockTestServer.java
CharlesDuboisSAP, line 9:
```suggestion
```

core/src/test/java/com/sap/ai/sdk/core/client/WireMockTestServer.java
```diff
@@ -16,12 +18,14 @@ abstract class WireMockTestServer {
...
   static WireMockServer wireMockServer;
   static Destination destination;
+  static ApiClient client;
 
   @BeforeAll
   static void setup() {
     wireMockServer = new WireMockServer(WIREMOCK_CONFIGURATION);
     wireMockServer.start();
     destination = DefaultHttpDestination.builder(wireMockServer.baseUrl()).build();
+    client = new AiCoreService().withDestination(destination).client();
```
CharlesDuboisSAP, line 28:
```suggestion
  static ApiClient client;

  @BeforeAll
  static void setup() {
    wireMockServer = new WireMockServer(WIREMOCK_CONFIGURATION);
    wireMockServer.start();
    final var destination = DefaultHttpDestination.builder(wireMockServer.baseUrl()).build();
    client = new AiCoreService().withDestination(destination).client();
```

core/src/test/java/com/sap/ai/sdk/core/AiCoreServiceTest.java
```diff
@@ -0,0 +1,143 @@
...
+    // verification
+    assertThat(destination.get(DestinationProperty.URI)).contains("https://foo.bar/v2/");
+    assertThat(destination.get(DestinationProperty.AUTH_TYPE)).isEmpty();
+    assertThat(destination.get(DestinationProperty.NAME)).isEmpty();
+    assertThat(destination.get(AI_CLIENT_TYPE_KEY)).contains(AI_CLIENT_TYPE_VALUE);
+    assertThat(client.getBasePath()).isEqualTo("https://foo.bar/v2/");
+  }
+
+  @Test
+  void testDeployment() {
```
CharlesDuboisSAP, line 99: There is barely any difference between `testDeployment` and `testSimpleCase`

core/src/test/java/com/sap/ai/sdk/core/AiCoreServiceTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+          @Nonnull
+          @Override
+          protected ApiClient getApiClient(@Nonnull Destination destination) {
+            return new ApiClient().setBasePath("https://fizz.buzz").setUserAgent("SAP");
+          }
+        };
+
+    final var customServiceForDeployment =
+        customService.forDeployment("deployment").withResourceGroup("group");
```
CharlesDuboisSAP, line 135: verify that `withResourceGroup` works

core/src/main/java/com/sap/ai/sdk/core/AiCoreDestination.java
newtork: Currently in `main` branch

* method `destination()` is required:
  * explicitly by OpenAI integration
  * implicitly by `client()`
* method `client()` is required:
  * explicit by Orchestration and Core integration

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
newtork: Agree, but I would rather change this (semi-related) improvement in a dedicated PR.

core/src/main/java/com/sap/ai/sdk/core/AiCoreDeployment.java
newtork: This method could be useful when retroactively changing the base destination.
```java
new AiCoreService()
  .forDeploymentByScenario("orchestration")
  .withDestination( mockedDestination )
  .client();
```

But I understand this is not a realistic scenario currently, when the following remains possible:
```java
new AiCoreService()
  .withDestination( mockedDestination )
  .forDeploymentByScenario("orchestration")
  .client();
```

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
newtork: API differences

| AiCoreService           | AiCoreDeployment  |
|-------------------------|-------------------|
| forDeployment           | withResourceGroup |
| forDeploymentByModel    |                   |
| forDeploymentbyScenario |                   |
| withDestination |                   |


API commons as represented by interface `AiCoreDestination`:

| AiCoreService           | AiCoreDeployment  |
|-------------------------|-------------------|
| client() | client() |
| destination() | destination()                  |


I don't know how I can reduce the code while still keeping the differences and similarities.

core/src/main/java/com/sap/ai/sdk/core/AiCoreDeployment.java
newtork: I can't change it to non-public because the method `withResourceGroup(String)` will be inaccessible then. (Right?)

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
newtork: I changed it slightly different.

core/src/test/java/com/sap/ai/sdk/core/AiCoreServiceTest.java
newtork: No caching, yet:
* `client()` implicitly calls `destination()`
* `destination()` implicitly looks up service bindings.

core/src/test/java/com/sap/ai/sdk/core/AiCoreServiceTest.java
newtork: Added test assertion here
https://github.com/SAP/ai-sdk-java/pull/79/commits/957510986a5d5ef499d4fcc586a60e193a4e07ae

core/src/main/java/com/sap/ai/sdk/core/AiCoreDeployment.java
CharlesDuboisSAP: Nope, still accessible from the `OpenAiClient` for example

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
CharlesDuboisSAP: I will create a follow-up PR for this

core/src/main/java/com/sap/ai/sdk/core/AiCoreDeployment.java
CharlesDuboisSAP: I did it in my follow up PR

CharlesDuboisSAP (dismissed):
Please merge #53 into this PR before merging this PR

CharlesDuboisSAP (approved):

PR #82, opened 2024-10-02
Added OWASP Dependency check

CharlesDuboisSAP:
## Context

AI/ai-sdk-java-backlog#65.

This is the equivalent of blackduck. It scans our dependencies and compares it to the `National Vulnerability Database`.
This is done on every commit.

### Feature scope:
 
- [x] Added OWASP Dependency check to the Fosstars report
[Successful run](https://github.com/SAP/ai-sdk-java/actions/runs/11214901629)

### DIFF
--- a/.github/workflows/fosstars-report.yml
+++ b/.github/workflows/fosstars-report.yml
@@ -5,6 +5,7 @@ on:
     - cron: "0 0 * * *" # every day at midnight
 
 env:
+  MVN_MULTI_THREADED_ARGS: --batch-mode --no-transfer-progress --fail-at-end --show-version --threads 1C
   JAVA_VERSION: 17
 
 jobs:
@@ -21,6 +22,11 @@ jobs:
           distribution: "temurin"
           java-version: ${{ env.JAVA_VERSION }}
           cache: 'maven'
+
+      - name: "Build SDK"
+        run: |
+          MVN_ARGS="${{ env.MVN_MULTI_THREADED_ARGS }} clean install -DskipTests -DskipFormatting"
+          mvn $MVN_ARGS
 
       - name: "OWASP Dependency check"
         run: mvn org.owasp:dependency-check-maven:10.0.4:check -DnvdApiKey=$NVD_API_KEY -DfailBuildOnCVSS=7 -DskipProvidedScope=true -DsuppressionFile=.pipeline/dependency-check-suppression.xml


dependency-check-suppression.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+  <suppress>
+    <notes><![CDATA[This is a JS dependency.]]></notes>
+    <cve>CVE-2021-41251</cve>
```
newtork, line 5: (Major/Question)

How exactly is the JS dependency recognized?
If we are affected by this false-positive(?) then all other users of "owasp" will have a similar experience, right?

CharlesDuboisSAP: It linked me to a dependency called `cloud-sdk-js@core 1.0`. I guessed it saw our openAI module use `com.sap.ai.sdk:core:0.1.0-SNAPSHOT` and thought it was good enough of a match...
I don't think this plugin is great, just better than blackduck hopefully.

.github/workflows/continuous-integration.yaml
```diff
@@ -90,6 +90,12 @@ jobs:
       - name: "Coverage Report"
         run: python .pipeline/scripts/print-coverage.py --jacoco-report-pattern "**/target/site/jacoco/jacoco.csv"
 
+      - name: "OWASP Dependency check"
+        run: mvn org.owasp:dependency-check-maven:10.0.4:check -DnvdApiKey=$NVD_API_KEY -DfailBuildOnCVSS=7 -DskipProvidedScope=true -DsuppressionFile=dependency-check-suppression.xml
```
newtork, line 94:
```suggestion
        run: mvn org.owasp:dependency-check-maven:10.0.4:check -DnvdApiKey=$NVD_API_KEY -DfailBuildOnCVSS=7 -DskipProvidedScope=true
```

I tested: Without the `-DsuppresionFile` the false-positive warnings are logged and the step is still successful.

I assume it would only fail with CVSS>=7. In that case I would suggest to completely remove the suppression file.

If you have a good reason to keep it, please move it to a subdirectory, not root path.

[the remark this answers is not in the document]
CharlesDuboisSAP: I will move it to the nightly `fosstars-report` workflow

.github/workflows/fosstars-report.yml
```diff
@@ -2,15 +2,51 @@ name: "Fosstars (Security)"
...
+        uses: actions/setup-java@v4
+        with:
+          distribution: "temurin"
+          java-version: ${{ env.JAVA_VERSION }}
+          cache: 'maven'
+
+      - name: "Build SDK"
+        run: |
+          MVN_ARGS="${{ env.MVN_MULTI_THREADED_ARGS }} clean install -DskipTests -DskipFormatting"
+          mvn $MVN_ARGS
```
newtork, line 29: (Comment)

`mvn install` is necessary, otherwise the dependency-check will fail on missing `core:0.1.0-SNAPSHOT`.

.github/workflows/fosstars-report.yml
```diff
@@ -2,15 +2,51 @@ name: "Fosstars (Security)"
...
+
+      - name: "OWASP Dependency check"
+        run: mvn org.owasp:dependency-check-maven:10.0.4:check -DnvdApiKey=$NVD_API_KEY -DfailBuildOnCVSS=7 -DskipProvidedScope=true -DsuppressionFile=.pipeline/dependency-check-suppression.xml
```
newtork, line 32: (Comment)

Be aware, having static plugin version in `yaml` (instead of `pom`) increases maintenance.

newtork (approved):

PR #101, opened 2024-10-17
chore: Include License in Release JAR Files

MatKuhr:
## Context

Required by OSPO


pom.xml
```diff
@@ -154,6 +154,19 @@
...
   </dependencies>
   <build>
+    <resources>
+      <resource>
+        <directory>${project.basedir}/src/main/resources</directory>
+      </resource>
+      <!--  include LICENSE in all jar files  -->
+      <resource>
+        <filtering>false</filtering>
+        <directory>${project.rootdir}</directory>
```
CharlesDuboisSAP, line 164:
```suggestion
        <directory>${user.dir}</directory>
```
That allows you to remove all `project.rootdir` properties

MatKuhr: This is not safe, because it depends from which directory the JVM was run. This might change e.g. if we make changes to how our GH actions work, or when running locally, etc. 

The problem is if this points to the wrong dir, we won't notice it because there is no error thrown or anything. So considering some options this seems to be the most reliable way. See also [this thread](https://stackoverflow.com/questions/3084629/finding-the-root-directory-of-a-multi-module-maven-reactor-project).

CharlesDuboisSAP (approved):

PR #103, opened 2024-10-17
Added resource group isolation to DeploymentCache

CharlesDuboisSAP:
## Context

AI/ai-sdk-java-backlog#54.

The deployment cache needs to isolate between resource groups.
Also customers should be able to reset the cache

### Feature scope:
 
- [x] Added resource group isolation to DeploymentCache
- [x] Added `AiCoreService.resetCache(client, resourceGroup)`

## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~

### DIFF
--- a/core/src/main/java/com/sap/ai/sdk/core/DeploymentCache.java
+++ b/core/src/main/java/com/sap/ai/sdk/core/DeploymentCache.java
@@ -4,12 +4,12 @@ import com.sap.ai.sdk.core.client.DeploymentApi;
 import com.sap.ai.sdk.core.client.model.AiDeployment;
 import com.sap.cloud.sdk.services.openapi.apiclient.ApiClient;
 import com.sap.cloud.sdk.services.openapi.core.OpenApiRequestException;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/sap/ai/sdk/core/DeploymentCache.java
```diff
@@ -107,25 +89,26 @@ private Optional<String> getDeploymentIdByModel(@Nonnull final String modelName)
...
                   .orElseThrow(
                       () ->
                           new NoSuchElementException(
                               "No running deployment found for scenario: " + scenarioId));
             });
   }
 
-  private Optional<String> getDeploymentIdByScenario(@Nonnull final String scenarioId) {
-    return CACHE.stream()
+  private Optional<String> getDeploymentIdByScenario(
```
MatKuhr, line 109: (minor) It would be nice to stick to either VAVRs `Option<>` or Javas `Optional`. Mixing the two is a bit awkward, and mixing them in public API might be confusing for users. In Cloud SDK we used VAVR because it's more powerful and integrated with `Try` etc.

CharlesDuboisSAP: This is not public API it's just a private helper function to reduce code duplication.

MatKuhr (approved):

PR #104, opened 2024-10-18
Lombok use of `val`

newtork:
- [x] core

* Generated code does not apply

### DIFF
--- a/core/src/test/java/com/sap/ai/sdk/core/client/DeploymentUnitTest.java
+++ b/core/src/test/java/com/sap/ai/sdk/core/client/DeploymentUnitTest.java
@@ -14,23 +14,17 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import com.sap.ai.sdk.core.client.model.AiDeployment;
 import com.sap.ai.sdk.core.client.model.AiDeploymentBulkModificationRequest;
-import com.sap.ai.sdk.core.client.model.AiDeploymentBulkModificationResponse;
 import com.sap.ai.sdk.core.client.model.AiDeploymentCreationRequest;
-import com.sap.ai.sdk.core.client.model.AiDeploymentCreationResponse;
-import com.sap.ai.sdk.core.client.model.AiDeploymentDeletionResponse;
-import com.sap.ai.sdk.core.client.model.AiDeploymentList;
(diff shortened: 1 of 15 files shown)


core/src/test/java/com/sap/ai/sdk/core/client/DeploymentUnitTest.java
```diff
@@ -14,23 +14,17 @@
...
-import com.sap.ai.sdk.core.client.model.AiDeploymentList;
 import com.sap.ai.sdk.core.client.model.AiDeploymentModificationRequest;
 import com.sap.ai.sdk.core.client.model.AiDeploymentModificationRequestWithIdentifier;
-import com.sap.ai.sdk.core.client.model.AiDeploymentModificationResponse;
-import com.sap.ai.sdk.core.client.model.AiDeploymentModificationResponseListInner;
 import com.sap.ai.sdk.core.client.model.AiDeploymentResponseWithDetails;
 import com.sap.ai.sdk.core.client.model.AiDeploymentStatus;
 import com.sap.ai.sdk.core.client.model.AiDeploymentTargetStatus;
 import com.sap.ai.sdk.core.client.model.AiExecutionStatus;
-import com.sap.ai.sdk.core.client.model.RTALogCommonResponse;
```
CharlesDuboisSAP, line 30: The explicit class names were an implicit test for breaking changes

newtork: Changes to class names result in *changes of file names*. This would be a better indicator for git-diff hopefully.

According to your logic, it would be best to have a Java file with **all** possible imports(?)

newtork:
Unfortunately PMD reports false-positvie.
[The issue was raised already in June, this year](https://github.com/pmd/pmd/issues/5079).

core/src/test/java/com/sap/ai/sdk/core/client/DeploymentUnitTest.java
MatKuhr: The problem with the git diff is that it's invisible between thousand of other changed files when re-generating. So I get the point, there is at least some usefulness to having compile checks.

On the other hand we can't guarantee api stability on the generated code, so it's also not critical. No strong opinion.

MatKuhr (approved):

PR #114, opened 2024-10-22
feat: Orchestration client

MatKuhr:
## Context

* AI/ai-sdk-java-backlog#55
* AI/ai-sdk-java-backlog#48

This PR introduces a completely new API for orchestration.

### Feature scope:
 
- [x] Dedicated client class
   - [x] Ready for Streaming
- [x] Streamlined config API via delegation
   - [x] Ready for Spring or Langchain integration
- [x] Convenience API for llm config
   - [x] Integrated with core class `AiModel`
- [x] Convenience API for messages
   - [x] ~Convenience for message history~ raised a feature request to enable this [here](https://github.tools.sap/AI/llm-orchestration/issues/662)
- [x] Convenience API for templating
- [x] Convenience API for filtering
- [x] Convenience API for masking

### TODO

- [x] Javadoc
- [x] Logging
- [ ] Mark generated models as `@Beta` and (maybe?) rename model package to `dto`
- [ ] Follow-up PR: Reduce code duplication with OpenAI client
- [ ] (maybe?) More tests asserting equals & hash code
- [ ] ...

## Definition of Done

- [ ] Aligned changes with the JavaScript SDK
- [ ] Documentation updated
- [ ] Release notes updated


orchestration/src/main/java/com/sap/ai/sdk/orchestration/ContentFilter.java
```diff
@@ -0,0 +1,3 @@
+package com.sap.ai.sdk.orchestration;
+
+public sealed interface ContentFilter permits AzureContentFilter {}
```
MatKuhr, line 3: Sealed is quite useful here, as this is a marker interface, and users can't supply their own implementations, because we wouldn't have the implementation to translate those to DTOs.

With a higher Java version, we could even switch over implementations of the interface. That has the nice benefit that the compiler can force us to handle all implementations. But only with Java 21 🤷🏻

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AzureContentFilter.java
```diff
@@ -0,0 +1,49 @@
...
+  public enum Sensitivity {
+    HIGH(0),
+    MEDIUM(2),
+    LOW(4);
+
+    private final int value;
+  }
+
+  @Nonnull
+  com.sap.ai.sdk.orchestration.client.model.FilterConfig toFilterConfigDTO() {
```
MatKuhr, line 31: I tried to consistently label the DTOs as such in the code, so that it's always obvious when we are (preparing) (de-) serialization

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationConfig.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Represents the configuration for the orchestration service. Allows for configuring the different
+ * modules of the orchestration service via a fluent API.
+ *
+ * @param <T> Type of the specific implementation to make a fluent API possible.
+ */
+public interface OrchestrationConfig<T extends OrchestrationConfig<T>> {
+
+  @Nonnull
+  Option<AiModel> getLlmConfig();
```
MatKuhr, line 16: Debatable, if we want to keep VAVR in our public API. Done here for now because this makes [our own code](https://github.com/SAP/ai-sdk-java/pull/114/files?diff=unified&w=0#diff-a3400feb9c493fbbc88bb89a699010093d0cb6a3364e016004ee417695746ff4R116) (arguably) much more beautiful 😄 

Alternatives would be Java `Optional` or just `@Nullable`.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationResponse.java
```diff
@@ -0,0 +1,49 @@
...
+import com.sap.ai.sdk.orchestration.client.model.CompletionPostResponse;
+import com.sap.ai.sdk.orchestration.client.model.TokenUsage;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import javax.annotation.Nonnull;
+import lombok.RequiredArgsConstructor;
+import lombok.val;
+
+public record OrchestrationResponse(
```
MatKuhr, line 13: Debatable if this should be a record or not. The AllArgs constructor is a bit awkward. Also, we might offer more convenience later on..

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationResponse.java
```diff
@@ -0,0 +1,49 @@
...
+    @Nonnull ChatMessage assistantMessage,
+    @Nonnull List<ChatMessage> allMessages,
```
MatKuhr, line 15: `allMessages` is a notable convenience for end users, this is not natively included in the orchestration response, but needed for chat scenarios or in general for an easy display of prompt + response.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
```diff
@@ -34,240 +23,88 @@
...
-   * template.
-   */
-  private static final Function<MaskingProviderConfig.MethodEnum, CompletionPostRequest>
-      MASKING_CONFIG =
-          (MaskingProviderConfig.MethodEnum maskingType) -> {
-            final var inputParams =
-                Map.of(
-                    "orgCV",
-                    """
-            Patrick Morgan
```
MatKuhr, line 118: While this is a nice example in principal, I felt that it is too large to properly see the relationship between masking config, prompt and result.

Also, I'm not sure if applying all masking profiles is a good idea.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AzureContentFilter.java
```diff
@@ -0,0 +1,49 @@
...
+@Accessors(fluent = true)
+public final class AzureContentFilter implements ContentFilter {
+  @Nullable private Sensitivity hate;
+  @Nullable private Sensitivity selfHarm;
+  @Nullable private Sensitivity sexual;
+  @Nullable private Sensitivity violence;
+
+  @RequiredArgsConstructor
+  public enum Sensitivity {
+    HIGH(0),
```
MatKuhr, line 23: Aligned w/ Orchestration team, to be aligned w/ JS and Launchpad

orchestration/pom.xml
```diff
@@ -36,32 +36,66 @@
...
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+
+    <!-- TODO: only needed for JsonObjectMapperBuilder, maybe we can use Jackson natively to avoid this dependency -->
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-web</artifactId>
+    </dependency>
```
newtork, line 93: (Comment)

I felt like I needed to intervene. But then I realized we also use this dependency in `openapi-core` from Cloud SDK :/

Might still be worthwhile to add a follow up / TGIF ticket

MatKuhr: Yeah, in `openapi-core` it is expected, because we use `RestTemplate`, but with this PR we no longer use `RestTemplate` for orchestration. So slightly different situation, but still worth a separate investigation outside of this PR.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AzureContentFilter.java
newtork: (Major)

Do you have a reference/link where this assignment 0=high is explained?
I would've assumed these are thresholds with terms high/low easily confusable. The last thing we want is adding confusion with those terms.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AzureContentFilter.java
newtork: I was under the impression that we camel-case abbreviations.

`toFilterConfigDto`

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationPrompt.java
```diff
@@ -0,0 +1,102 @@
...
+    this(List.of(new UserMessage(message)), Map.of());
+  }
+
+  /**
+   * Initialize a prompt from the given messages.
+   *
+   * @param message The first message.
+   * @param messages Optionally, more messages.
+   */
+  public OrchestrationPrompt(@Nonnull final Message message, @Nullable final Message... messages) {
```
newtork, line 54: (Minor)

AFAIK `@Nullable` varargs only makes sense if you expect an explicitly invoked `new OrchestrationPrompt(msg, (Message[]) null)`, which I wouldn't find acceptable. Empty varargs are handled as empty array. Please correct me if I'm wrong here.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationConfig.java
```diff
@@ -0,0 +1,136 @@
...
+ *   <li>LLM Config (Mandatory)
+ *   <li>Templating (Optional)
+ *   <li>Data Masking (Optional)
+ *   <li>Input Content Filtering (Optional)
+ *   <li>Output Content Filtering (Optional)
+ * </ul>
+ *
+ * @param <T> Type of the specific implementation to make a fluent API possible.
+ */
+public interface OrchestrationConfig<T extends OrchestrationConfig<T>> {
```
newtork, line 25: (Comment)

image

The set of available constructors look slightly confusing/inconsistent to me.

If we wanted to change that, then I would suggest (1) static factory methods. And if I understand correctly there are two distinct creation paths for this class, maybe it makes sense to (2) derive sub-classes for message-based prompts and template-based prompts.

Sub classes of (2) could be returned by factory methods of (1). Thus API remains explorable.

orchestration/src/test/java/com/sap/ai/sdk/orchestration/ModuleConfigFactoryTest.java
```diff
@@ -0,0 +1,176 @@
...
+  void testEmptyFilter() {
+    var inputFilter = new AzureContentFilter();
+    assertThatThrownBy(inputFilter::toFilterConfigDTO)
+        .isInstanceOf(IllegalStateException.class)
+        .hasMessageContaining(
+            "When configuring an azure content filter, at least one filter category must be set");
+  }
+
+  @Test
+  void testMasking() {
```
CharlesDuboisSAP, line 163: Add pseudonmization testing

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationPrompt.java
MatKuhr: You are completely right, `@Nullable` doesn't make sense. Fixed it here and also in other places I used this pattern.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AzureContentFilter.java
MatKuhr: true, fixed

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
```diff
@@ -35,239 +24,86 @@
...
-                    OrchestrationConfig.create()
-                        .moduleConfigurations(
-                            ModuleConfigs.create()
-                                .llmModuleConfig(LLM_CONFIG)
-                                .templatingModuleConfig(templatingConfig)
-                                .maskingModuleConfig(maskingConfig)))
-                .inputParams(inputParams);
-          };
+  private final OrchestrationClient client =
+      new OrchestrationClient().withLlmConfig(OpenAiModel.GPT_35_TURBO);
```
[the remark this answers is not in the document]
MatKuhr: While we could, the orchestration `LlmConfig` not only contains the model, but also optionally hyperparameters. In OpenAI, the hyperparameters are separate. Also, `forModel` currently does a deployment ID lookup and only creates the instance afterwards, where orchestration does no such lookup.

So I'd feel its different enough to deserve somewhat different terminology, even though there are similarities. No strong opinion though.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AzureContentFilter.java
MatKuhr: The terminology isn't consistent yet, which is why this is so confusing, which is why this enum exists. Currently, the [orchestration docs](https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/orchestration-workflow) still use largely numbers, AI Launchpad and the Python SDK use the terms low / medium / high, but with the inverse semantic of high being "high threshold" i.e. allow everything.

Aligning on a consistent terminology is out of scope for this PR, will be done separately. So for now maybe consider the current state a placeholder for whatever will be aligned. Or maybe we should completely remove it from this PR and switch back to plain numbers, until alignment is done. No strong opinion.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
```diff
@@ -0,0 +1,200 @@
+package com.sap.ai.sdk.orchestration;
```
CharlesDuboisSAP, line 1:
```suggestion
package com.sap.ai.sdk.orchestration.client;
```
Why is the package not `client`?

orchestration/pom.xml
```diff
@@ -74,54 +108,31 @@
...
-            <id>delete-orchestration-generated-client</id>
-          </execution>
-        </executions>
-      </plugin>
       <plugin>
         <groupId>com.sap.cloud.sdk.datamodel</groupId>
         <artifactId>openapi-generator-maven-plugin</artifactId>
         <configuration>
+          <skip>true</skip>
+          <!-- skip automatic generation until we can omit API classes from code generation -->
```
CharlesDuboisSAP, line 130: We need a BLI for this, Junjie needs to be aware

orchestration/src/test/java/com/sap/ai/sdk/orchestration/SerializationTest.java
```diff
@@ -0,0 +1,139 @@
...
+                    .inputMasking(inputMaskingResult)
+                    .outputUnmasking(outputMaskingResult))
+            .orchestrationResult(orchestrationResult);
+
+    var actual =
+        OrchestrationClient.JACKSON.readValue(
+            getClass().getClassLoader().getResourceAsStream("serializedResponse.json"),
+            CompletionPostResponse.class);
+
+    assertThat(actual).isEqualTo(expected);
```
CharlesDuboisSAP, line 137: This test needs to better reflect exceptions and stacktraces that the customer would encounter with a malformed response. Currently `JACKSON.readValue()` never throws with a malformed response.

PR #111, opened 2024-10-21
Run Checkstyle by default

newtork:
- [x] Make the code quality related plugins runnable from sub modules.
- [x] Enable Checkstyle by default

Sample output:
```
mvn clean install -f foundationmodels/openai

[...]
[INFO] 
[INFO] --- checkstyle:3.5.0:check (default) @ openai ---
[INFO] There are 3 errors reported by Checkstyle 10.18.2 with C:\SAPDevelop\ai-sdk-java\foundation-models\openai/../..//.pipeline/checkstyle.xml ruleset.
[ERROR] src\main\java\com\sap\ai\sdk\foundationmodels\openai\model\OpenAiChatCompletionParameters.java:[95,39] (misc) TodoComment: Comment matches to-do format '(TODO)|(todo)|(ToDo)|(FIXME)|(fixme)|(FixMe)'.
[ERROR] src\main\java\com\sap\ai\sdk\foundationmodels\openai\model\OpenAiChatMessage.java:[255,7] (misc) TodoComment: Comment matches to-do format '(TODO)|(todo)|(ToDo)|(FIXME)|(fixme)|(FixMe)'.
[ERROR] src\main\java\com\sap\ai\sdk\foundationmodels\openai\model\OpenAiChatMessage.java:[271,11] (misc) TodoComment: Comment matches to-do format '(TODO)|(todo)|(ToDo)|(FIXME)|(fixme)|(FixMe)'.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  10.397 s
[INFO] Finished at: 2024-10-21T10:48:31+02:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.5.0:check (default) on project openai: You have 3 Checkstyle violations. -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
```

PR #106, opened 2024-10-18
Run PMD by default

newtork:
- [x] Make the code quality related plugins runnable from sub modules.
- [x] Enable PMD by default

Sample output:
```
mvn clean install -f core

[...]

[INFO] --- pmd:3.25.0:pmd (pmd) @ core ---
[INFO] PMD version: 7.3.0
[INFO] Rendering content with org.apache.maven.skins:maven-fluido-skin:jar:2.0.0-M9 skin
[INFO] 
[INFO] <<< pmd:3.25.0:check (default) < :pmd @ core <<<
[INFO] 
[INFO] 
[INFO] --- pmd:3.25.0:check (default) @ core ---
[WARNING] PMD Failure: com.sap.ai.sdk.core.AiCoreDestination:25 Rule:LocalVariableCouldBeFinal Priority:3 Local variable 'destination' could be declared final.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  21.686 s
[INFO] Finished at: 2024-10-18T12:45:48+02:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-pmd-plugin:3.25.0:check (default) on project core: PMD 7.3.0 has found 1 violation. For more details see: C:\SAPDevelop\ai-sdk-java\core\target\pmd.xml -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
```

PR #115, opened 2024-10-24
[CI/CD] Remove redundant steps from workflow: Checkstyle, PMD, Spotbugs 

newtork:
The following steps are being invoked by default with every `mvn clean install` already:
* [x] [Checkstyle](https://github.com/SAP/ai-sdk-java/pull/111)
* [x] [PMD](https://github.com/SAP/ai-sdk-java/pull/106)
* [x] [Spotbugs](https://github.com/SAP/ai-sdk-java/pull/109)

(They are listening on maven phase `verify`.)


newtork:
> The printing is still useful for some of these

When the preceding build step already fails having `clean install` then the subsequent steps will not be executed.

> In particular since they attach it to the action result which can be easier to read than logs?

Do you mean the workflow summary? The python scripts wrote to them.

While I understand the concern that `mvn` log is difficult to handle, currently **any** metric >0 will not be tolerated.
Therefore I'm not sure what benefit such table brings, beyond much better format. The information itself remains the same.

MatKuhr:
Agreed, we could add a `failure()` conditional to the printing, but if the mvn output is readable enough, it's not needed.

CharlesDuboisSAP (approved):

CharlesDuboisSAP (approved):

PR #123, opened 2024-10-28
[DevOps] Skip Code Generation By Default

MatKuhr:
## Context

This speeds up the build from 11s down to 1s on my machine when running `mvn compile`. For `mvn clean compile` the gain is still around 5 seconds.


CharlesDuboisSAP (commented):
Add `-Drelease` to the [weekly automated spec updates](https://github.tools.sap/cloudsdk/cloud-sdk-java-tests/actions)

MatKuhr:
> Add `-Drelease` to the [weekly automated spec updates](https://github.tools.sap/cloudsdk/cloud-sdk-java-tests/actions)

✅ [Done](https://github.tools.sap/cloudsdk/cloud-sdk-java-tests/commits/main)

core/pom.xml
```diff
@@ -142,62 +142,53 @@
...
-              </includes>
-            </fileset>
-          </filesets>
-        </configuration>
-        <executions>
-          <execution>
-            <id>delete-aicore-generated-client</id>
-          </execution>
-        </executions>
-      </plugin>
```
CharlesDuboisSAP, line 164: The delete-aicore-generated-client step is missing in the new profile.
Same for the orchestration client but it should skip, because generation is disabled.

CharlesDuboisSAP (changes requested):

core/pom.xml
MatKuhr: It's not needed, the deletion happens as part of the genereation, I enabled the flag. It only deletes packages now, not the full directory, so it doesn't delete the custom code

.github/workflows/continuous-integration.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
 
       - name: "Build SDK"
         run: |
-          MVN_ARGS="${{ env.MVN_MULTI_THREADED_ARGS }} clean install -DskipTests -DskipFormatting"
+          MVN_ARGS="${{ env.MVN_MULTI_THREADED_ARGS }} clean install -Dgenerate -DskipTests -DskipFormatting"
```
newtork, line 53: (Question)

Just to clarify, the generation happens with every commit of every PR right?

MatKuhr: Yes 👍🏻

PR #124, opened 2024-10-28
Orchestration client suggestions

CharlesDuboisSAP:
My comments on Matthias' PR

### DIFF
--- a/orchestration/src/main/java/com/sap/ai/sdk/orchestration/ModuleConfigFactory.java
+++ b/orchestration/src/main/java/com/sap/ai/sdk/orchestration/ModuleConfigFactory.java
@@ -32,7 +32,7 @@ final class ModuleConfigFactory {
             .map(ModuleConfigFactory::toLlmModuleConfigDTO)
             .getOrElseThrow(() -> new IllegalStateException("LLM module config is required"));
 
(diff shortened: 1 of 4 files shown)


orchestration/src/main/java/com/sap/ai/sdk/orchestration/ModuleConfigFactory.java
```diff
@@ -32,7 +32,7 @@ static ModuleConfigs toModuleConfigDTO(
             .map(ModuleConfigFactory::toLlmModuleConfigDTO)
             .getOrElseThrow(() -> new IllegalStateException("LLM module config is required"));
 
-    val template = config.getTemplate().getOrElse(() -> TemplateConfig.fromMessages(List.of()));
```
MatKuhr, line 35: Is this the only motivation for the public API change? Because for regular users of the public API, at least one message is required for this to make sense.

Here I am just abusing the API internally, passing an empty list, to make the internal code look slightly more sleek.

CharlesDuboisSAP: My motivation was minimal public API

PR #125, opened 2024-10-28
Chore/sample app readme update

MatKuhr:
Minor improvement based on user feedback

### DIFF
--- a/sample-code/spring-app/README.md
+++ b/sample-code/spring-app/README.md
@@ -5,15 +5,25 @@
 Sample code to demonstrate the usage of the SAP AI SDK.
 Also used as basis for running E2E tests.
 
-## Build, Run, Deploy Locally
+## Build and Run the Sample App
 
-Build the project with:
+Before you can run the sample app, you need to install the AI SDK into your local Maven repository:
 
-1. `mvn compile`
-2. [Download a service key for your AI Core service instance](../../README.md#set-credentials-as-dedicated-environment-variable)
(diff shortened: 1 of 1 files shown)


sample-code/spring-app/README.md
```diff
@@ -5,15 +5,25 @@
...
-1. `mvn compile`
-2. [Download a service key for your AI Core service instance](../../README.md#set-credentials-as-dedicated-environment-variable)
-3. Create the environment variable `AICORE_SERVICE_KEY`
-4. Run the application with `mvn spring-boot:run`
-5. [See all available endpoints](localhost:8080)
+* Run `mvn install -DskipTests` from the root directory of the repository.
+
+> [!NOTE]  
+> The sample app uses the latest state of the SDK, so make sure to install the SDK after pulling a new version via Git.
+> Alternatively, you check out one of the release tags of the repository, e.g. `git fetch --all --tags && git checkout rel/0.1.0`.
```
newtork, line 16: (Minor)

Not a fan of this/a note. Semantically it's correct, but **visually** it will be most prominent, highlighted paragraph of the document.

newtork (approved):

sample-code/spring-app/README.md
MatKuhr: Yeah, but I feel that visually it's not as prominent as the notes on Docusaurus, because it doesn't have a different background, it's "just" a blue bar at the right. So maybe still acceptable..

PR #127, opened 2024-10-29
Simpler Config Variant

MatKuhr:
Alternative API to #114 

Differences:

* Full config now has to be provided with each request
* Config may be stored, it's (shallow) immutable 
* New configs can easily be created from a base config per request using the "with" methods
* Implementation is much simpler


orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationConfig.java
```diff
@@ -1,136 +1,29 @@
...
-   * @param filter The content filter configuration to be applied to output
-   * @return Fluent API object of this config instance
-   * @see #withInputContentFilter(ContentFilter)
-   * @see ContentFilter
-   * @see AzureContentFilter
-   */
-  @Nonnull
-  T withOutputContentFilter(@Nonnull final ContentFilter filter);
+@Value
+@With
```
MatKuhr, line 16: Making the config immutable allows us to use Lomboks `@With` here, shortcutting implementing the `with...` methods

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
```diff
@@ -24,8 +26,9 @@
 @RequestMapping("/orchestration")
 class OrchestrationController {
 
-  private final OrchestrationClient client =
-      new OrchestrationClient().withLlmConfig(OpenAiModel.GPT_35_TURBO);
+  private final OrchestrationClient client = new OrchestrationClient();
+  private final OrchestrationConfig config =
+      new OrchestrationConfig().withLlmConfig(OpenAiModel.GPT_35_TURBO);
```
MatKuhr, line 31: users have to maintain two objects now

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationConfig.java
```diff
@@ -1,136 +1,29 @@
...
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class OrchestrationConfig {
+
+  @Nullable AiModel llmConfig;
+  @Nullable TemplateConfig template;
+  @Nullable MaskingConfig maskingConfig;
+  @Nullable ContentFilter inputContentFilter;
+  @Nullable ContentFilter outputContentFilter;
+
+  public OrchestrationConfig() {
```
MatKuhr, line 26: Slight variation:

```suggestion
  public OrchestrationConfig(AiModel llmConfig) {
```

Upside: Users see that this is required, downside: users have to provide it initially, even if they want to change it per request.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
```diff
@@ -82,9 +74,10 @@ public OrchestrationClient(@Nonnull final AiCoreDeployment deployment) {
    * @throws OrchestrationClientException if the request fails
    */
   @Nonnull
-  public String chatCompletion(@Nonnull final String userPrompt)
+  public String chatCompletion(
+      @Nonnull final String userPrompt, @Nonnull final OrchestrationConfig config)
```
CharlesDuboisSAP, line 78: I would have `OrchestrationClient.withConfig(config).chatCompletion("prompt")`

CharlesDuboisSAP (commented):
I love this, simpler is better. I also want @rpanackal 's review.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
MatKuhr: But then what would be the result of `OrchestrationClient.withConfig(config)` alone? A `PromptBuilder` or something like that?

CharlesDuboisSAP: The config is stored on the client and `withConfig()` returns `this`

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
```diff
@@ -50,8 +53,10 @@ OrchestrationResponse completion() {
   OrchestrationResponse systemMessage() {
     final var prompt =
         new OrchestrationPrompt(
-            new SystemMessage("Show the user your superior geographical knowledge!"),
-            new UserMessage("What is the capital of France?"));
+            List.of(
+                new SystemMessage("Show the user your superior geographical knowledge!"),
+                new UserMessage("What is the capital of France?")),
+            config);
```
newtork, line 59: (Nitpick)

There goes the powerpoint readiness ;)

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
```diff
@@ -71,8 +76,8 @@ OrchestrationResponse template() {
     final var inputParams = Map.ofEntries(templateVariable.apply("german"));
 
     final var prompt =
-        new OrchestrationPrompt(inputParams)
-            .withTemplate(TemplateConfig.fromMessages(templateMessage));
+        new OrchestrationPrompt(
+            inputParams, config.withTemplate(TemplateConfig.fromMessages(templateMessage)));
 
     return client.chatCompletion(prompt);
```
newtork, line 82: (Question)

Have you considered moving the config argument to `#chatCompletion(` instead of `new OrchestrationPrompt(`?
(This would reduce the object nesting by 1 level)

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
jjtang1985: Maybe moving messages to the second place, so we can use varargs?

OrchestrationPrompt(OrchestrationConfig config, Messages... messages)

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
```diff
@@ -83,8 +88,8 @@ OrchestrationResponse filter(@Nonnull @PathVariable(name = "level") final Sensit
     final var filter = new AzureContentFilter().hate(level);
     final var prompt =
         new OrchestrationPrompt(
-                "This prompt demonstrates how to hit the fucking input filter. And hit it hard, like we mean it.")
-            .withInputContentFilter(filter);
+            "This prompt demonstrates how to hit the fucking input filter. And hit it hard, like we mean it.",
+            config.withInputContentFilter(filter));
```
jjtang1985, line 92: I'm not sure whether a "prompt" should know the orchestration config, or should the prompt be a message.

How about the following?
```
// update config
config.withInputContentFilter(filter));
// init a prompt, being a message
final var prompt = new OrchestrationPrompt("my prompt");
// send a request with a prompt + a config
return client.chatCompletion(prompt, config);
```

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
jjtang1985: I guess there must be good reasons to split the client and config.
Eg., you can reuse the client with different config.
If so, keeping two objects is fine.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
```diff
@@ -71,8 +76,8 @@ OrchestrationResponse template() {
     final var inputParams = Map.ofEntries(templateVariable.apply("german"));
 
     final var prompt =
-        new OrchestrationPrompt(inputParams)
-            .withTemplate(TemplateConfig.fromMessages(templateMessage));
+        new OrchestrationPrompt(
+            inputParams, config.withTemplate(TemplateConfig.fromMessages(templateMessage)));
```
jjtang1985, line 80: I started writing this message after my previous [comment](https://github.com/SAP/ai-sdk-java/pull/127/files#r1827819209).

I feel an OrchestrationPrompt is another config that knows prompt templates in this case.

```
// update config
config.withTemplate(TemplateConfig.fromMessages(templateMessage));
// send a request with an inputParams + a config
return client.chatCompletion(inputParams, config);
```

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
MatKuhr: You can use the same client with different config objects, but the reason this is now two objects instead of one is that the implementation is simpler

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
MatKuhr: Fixed 👍🏻

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
MatKuhr: Implemented this now, I think it's probably slightly better

rpanackal:
I like the new approach. Its more deliberate handling of config for users but straightforward.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
jjtang1985: > It's somewhat funny that you are suggesting an API usage that would exactly not work 😄

You got my point 😄  which is similar to Alex's suggestion.

PR #136, opened 2024-11-04
Reduce visibility AiCoreService#DOTENV from public to private

newtork:
Exposing this field publicly leads to some risks:
* User expects API stability
* .. for a field with a third-party class
* Implicit invitation for more util fields (feature creep)
* Auto-completion may confuse the user

### DIFF
--- a/core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
+++ b/core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
@@ -43,7 +43,7 @@ public class AiCoreService implements AiCoreDestination {
   private static final String AI_RESOURCE_GROUP = "URL.headers.AI-Resource-Group";
 
   /** loads the .env file from the root of the project */
-  protected static final Dotenv DOTENV = Dotenv.configure().ignoreIfMissing().load();
+  private static final Dotenv DOTENV = Dotenv.configure().ignoreIfMissing().load();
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
```diff
@@ -43,7 +43,7 @@ public class AiCoreService implements AiCoreDestination {
   private static final String AI_RESOURCE_GROUP = "URL.headers.AI-Resource-Group";
 
   /** loads the .env file from the root of the project */
-  public static final Dotenv DOTENV = Dotenv.configure().ignoreIfMissing().load();
+  protected static final Dotenv DOTENV = Dotenv.configure().ignoreIfMissing().load();
```
CharlesDuboisSAP, line 46:
```suggestion
  private static final Dotenv DOTENV = Dotenv.configure().ignoreIfMissing().load();
```
Why not private?

CharlesDuboisSAP (commented):
my bad

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
newtork: > Why not private?

`private` would also be fine for me.
I consider the possibility for users to extend/customize the class. It may be useful to them to load environment variables too. Not a strong opinion.

CharlesDuboisSAP (approved):

PR #139, opened 2024-11-05
Added dependencies instructions

CharlesDuboisSAP:
## Context

A customer was overriding the Cloud SDK version to an older one. This is due to the fact that they are using Cloud SDK in addition to the AI SDK

> ⚠️ Cloud SDK users should remove version tags from all Cloud SDK dependencies.


newtork (changes requested):
I don't understand why we should solve/document dependency convergence for users. Our markdown files (and similarly the whole Cloud SDK documentation portal) are suffering from these notes, warnings and information boxes.

CharlesDuboisSAP:
> I don't understand why we should solve/document dependency convergence for users.

Because most of our customers will be existing users of Cloud SDK and we know that they can't set their dependencies correctly

PR #145, opened 2024-11-08
Removed mentions of "DTO"

CharlesDuboisSAP:


orchestration/src/main/java/com/sap/ai/sdk/orchestration/ConfigToRequestTransformer.java
```diff
@@ -12,28 +13,27 @@
 import lombok.NoArgsConstructor;
 import lombok.val;
 
-/** Factory to create all DTOs from an orchestration configuration. */
+/** Factory to create all data objects from an orchestration configuration. */
 @NoArgsConstructor(access = AccessLevel.NONE)
-final class ModuleConfigFactory {
+final class ConfigToRequestTransformer {
```
CharlesDuboisSAP, line 18: Up for discussion, as long as it's not factory or DTO

PR #146, opened 2024-11-08
Openapi generator orchestration

rpanackal:
## Context

AI/ai-sdk-java-backlog#99.

Support oneOf from Orchestration spec using OpenAPI code generator. 
Avoid template and spec modification.

### Feature scope:
 
- [x] Generate model classes only
- [x] Freeze Interfaces generated, or rather avoid overwrite
- [x] Improved annotation of interfaces: Deserialisation by existence of properties. 
- [x] Introduce property type based implementation resolver for oneOfs.

## Definition of Done

- [x] ~Documentation updated~
- [x] ~Release notes updated~


orchestration/src/main/java/com/sap/ai/sdk/orchestration/client/model/MaskingProviderConfig.java
```diff
@@ -12,395 +12,17 @@
...
+    value = "org.openapitools.codegen.languages.JavaClientCodegen",
+    date = "2024-11-08T10:43:10.628772+01:00[Europe/Berlin]",
+    comments = "Generator version: 7.9.0")
+@JsonIgnoreProperties(
+    value = "", // ignore manually set , it will be automatically generated by Jackson during
+    // serialization
+    allowSetters = true // allows the  to be set during deserialization
+    )
+@JsonTypeInfo(use = JsonTypeInfo.Id.DEDUCTION)
+public interface MaskingProviderConfig {}
```
newtork, line 28: (Question)

This is custom / hand-crafted code(?)
How do you protect this class from being overwritten when the generator runs next time?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/GeneralizedFallbackDeserializer.java
```diff
@@ -0,0 +1,73 @@
...
+ * <p>This deserializer is useful when the exact subtype of an object is not known in advance, but
+ * the possible subtypes are known and annotated with {@link JsonSubTypes}.
+ *
+ * <p>When deserializing, this deserializer will attempt to deserialize the JSON object into each of
+ * the possible subtypes in turn. The first successful deserialization will be returned. If none of
+ * the deserializations are successful, a {@link JsonMappingException} will be thrown.
+ *
+ * @param <T> The base type to deserialize into.
+ */
+public class GeneralizedFallbackDeserializer<T> extends StdDeserializer<T> {
```
newtork, line 22: (Question)

In what cases do we need this class? I would've expected this was default Jackson behavior 🤔

orchestration/src/main/java/com/sap/ai/sdk/orchestration/LLMModuleResultDeserializer.java
```diff
@@ -0,0 +1,77 @@
...
+import com.sap.ai.sdk.orchestration.client.model.LLMModuleResult;
+import com.sap.ai.sdk.orchestration.client.model.LLMModuleResultStreaming;
+import com.sap.ai.sdk.orchestration.client.model.LLMModuleResultSynchronous;
+import java.io.IOException;
+
+/**
+ * A deserializer for {@link LLMModuleResult} that determines the concrete implementation based on
+ * the structure of the JSON object.
+ */
+public class LLMModuleResultDeserializer extends StdDeserializer<LLMModuleResult> {
```
newtork, line 15: (Question)

Do we need this class?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/client/model/LLMModuleResult.java
```diff
@@ -12,470 +12,15 @@
...
+
+// ignore manually set , it will be automatically generated by Jackson during
+// serialization
+// allows the  to be set during deserialization
+@JsonTypeInfo(use = JsonTypeInfo.Id.NONE)
+@JsonSubTypes({
+  @JsonSubTypes.Type(value = LLMModuleResultStreaming.class, name = "LLMModuleResultStreaming"),
+  @JsonSubTypes.Type(value = LLMModuleResultSynchronous.class, name = "LLMModuleResultSynchronous")
+})
+public interface LLMModuleResult {}
```
newtork, line 26: (Question)

This is hand-crafted right?
Why not use deduction?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/client/model/MaskingProviderConfig.java
newtork: > This is custom / hand-crafted code(?)

_"All interface types are hand-crafted"_

> How do you protect this class from being overwritten when the generator runs next time?

_"see openapi generator ignore file"_

orchestration/src/main/java/com/sap/ai/sdk/orchestration/LLMModuleResultDeserializer.java
newtork: > Do we need this class?

_"Deduction does not look recursively for discriminating field names. But that would be possible here."_

orchestration/src/main/java/com/sap/ai/sdk/orchestration/client/model/LLMModuleResult.java
newtork: > This is hand-crafted right?

_"yes"_

> Why not use deduction?

_"Deduction does not look recursively for discriminating field names. But that would be possible here."_

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
```diff
@@ -40,6 +41,8 @@ public class OrchestrationClient {
             .visibility(PropertyAccessor.GETTER, JsonAutoDetect.Visibility.NONE)
             .visibility(PropertyAccessor.SETTER, JsonAutoDetect.Visibility.NONE)
             .serializationInclusion(JsonInclude.Include.NON_NULL)
+            .deserializerByType(
+                LLMModuleResult.class, new GeneralizedFallbackDeserializer<>(LLMModuleResult.class))
```
newtork, line 45: (Major)

Please decide on which Deserializer to use for `LLMModuleResult`. Your PR ships two classes, one unused. Let's avoid "dead" code.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/LLMModuleResultDeserializer.java
```diff
@@ -0,0 +1,77 @@
...
+              ctxt.findRootValueDeserializer(ctxt.constructType(concreteClass));
+
+          return (LLMModuleResult) deserializer.deserialize(rootParser, ctxt);
+        }
+      }
+    }
+
+    // If unable to determine, throw an exception or handle default case
+    throw new JsonMappingException(
+        p, "Unable to determine the concrete implementation of LLMModuleResult");
```
newtork, line 75: (Major)

Please check whether the following cases are handled correctly:

```
choices = [{ }]    // fine, handled
choices = [   ]    // error
choices = null     // error
choices = "string" // error
```

I think (2) and (3) should not necessarily throw an exception(?). May be discussion worthy.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/client/model/FilterConfig.java
```diff
@@ -12,249 +12,17 @@
...
+
+@javax.annotation.Generated(
+    value = "org.openapitools.codegen.languages.JavaClientCodegen",
+    date = "2024-11-08T10:43:10.628772+01:00[Europe/Berlin]",
+    comments = "Generator version: 7.9.0")
+@JsonIgnoreProperties(
+    value = "", // ignore manually set , it will be automatically generated by Jackson during
+    // serialization
+    allowSetters = true // allows the  to be set during deserialization
+    )
```
newtork, line 26: (Minor)

Maybe we can remove this. To not give wrong impression about this class being generated.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/GeneralizedFallbackDeserializer.java
rpanackal: I have removed `GeneralizedFallbackDeserializer`.

Jackson deserialiser resolves for sub type based on 
1. existence of any unique property by name 
2. values of some specified string property

For `LLMModuleResult` oneof interface, neither approaches can help resolve to one of the concrete classes, as the only discriminating information is the type of one of their properties. Similar cases seems highly likely unless OpenAPI spec definition takes this into account.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/LLMModuleResultDeserializer.java
rpanackal: No. We finalised to keep this class instead of `GeneralizedFallbackDeserializer` to handle oneOf for `LLMModuleResult`

orchestration/src/main/java/com/sap/ai/sdk/orchestration/LLMModuleResultDeserializer.java
rpanackal: (2) and (3) is also the case where we simply have no information to infer the subtype. So, I assume it makes sense to raise a `JsonMappingException`.

newtork: > another way of handling this.

As a user my expectation would be:
```
choices = [{ }]    // -> non-empty List
choices = [   ]    //  -> empty List
choices = null     // -> null
choices = "string" // error
```

Or do I miss something?

orchestration/pom.xml
```diff
@@ -118,54 +118,73 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <!-- Additional dependencies for OSS code generator resttemplate  -->
+    <dependency>
+      <groupId>org.openapitools</groupId>
+      <artifactId>jackson-databind-nullable</artifactId>
```
newtork, line 125: (Question)

What do we need this dependency for?

orchestration/pom.xml
```diff
@@ -118,54 +118,73 @@
...
+    </dependency>
+    <dependency>
+      <groupId>javax.annotation</groupId>
+      <artifactId>javax.annotation-api</artifactId>
```
newtork, line 129: (Question)

What do we need this dependency for?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/LLMModuleResultDeserializer.java
```diff
@@ -0,0 +1,82 @@
...
+   * @param context The deserialization context.
+   * @param concreteType The concrete type to deserialize into.
+   * @return The deserialized object.
+   * @throws IOException If an I/O error occurs.
+   */
+  private LLMModuleResult delegateToDefaultDeserializer(
+      JsonParser parser, DeserializationContext context, JavaType concreteType) throws IOException {
+    var defaultDeserializer = context.findRootValueDeserializer(concreteType);
+    return (LLMModuleResult) defaultDeserializer.deserialize(parser, context);
+  }
```
newtork, line 81: (Comment)

I'm okay for the extra method, we can keep it.

But just a little remark:
* Since technically, this is a two-liner, without multiple re-use, I would probably just move it into the `deserialize` method above. In some cases it still makes sense to decouple a few lines into dedicated methods: e.g. dedicated method testing or method customization via class extension and overriding. But I wouldn't see it here.
* JavaDoc on private methods is appreciated but not required. Personally I rather prefer @Nullable / @Nonnull - optional on private methods as well.
* Method could be static.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/NoTypeInfoMixin.java
```diff
@@ -0,0 +1,7 @@
+package com.sap.ai.sdk.orchestration;
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+/** Mixin to suppress @JsonTypeInfo for oneOf interfaces. */
+@JsonTypeInfo(use = JsonTypeInfo.Id.NONE)
+public interface NoTypeInfoMixin {}
```
newtork, line 7: (Question/Minor)

Can this class be package-private?
If not, could we append `@Beta` annotation?

orchestration/src/test/java/com/sap/ai/sdk/orchestration/LLMModuleResultDeserializerTest.java
```diff
@@ -0,0 +1,95 @@
...
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.sap.ai.sdk.orchestration.client.model.LLMModuleResult;
+import com.sap.ai.sdk.orchestration.client.model.LLMModuleResultStreaming;
+import com.sap.ai.sdk.orchestration.client.model.LLMModuleResultSynchronous;
+import lombok.SneakyThrows;
+import org.junit.jupiter.api.Test;
+
+class LLMModuleResultDeserializerTest {
```
newtork, line 12: (Comment/Minor)

Test class looks good!
Only wondering whether we could remove a few new-lines. Or remove the comments.

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/OrchestrationTest.java
```diff
@@ -25,7 +27,12 @@ void testCompletion() {
     final var result = controller.completion();
 
     assertThat(result).isNotNull();
-    assertThat(result.getOrchestrationResult().getChoices().get(0).getMessage().getContent())
+    assertThat(
+            ((LLMModuleResultSynchronous) result.getOrchestrationResult())
+                .getChoices()
+                .get(0)
+                .getMessage()
+                .getContent())
```
newtork, line 35: (Question/Major)

If I understand correctly, the PR changes public API in a way, that requires users to cast a result object to a specific class, before being able to read the response content. Is that correct? Is that the expected result?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/client/model/SearchDocumentKeyValueListPair.java
```diff
@@ -12,93 +12,68 @@
...
-  @JsonProperty("key")
+@JsonPropertyOrder({
+  SearchDocumentKeyValueListPair.JSON_PROPERTY_KEY,
+  SearchDocumentKeyValueListPair.JSON_PROPERTY_VALUE,
+  SearchDocumentKeyValueListPair.JSON_PROPERTY_SELECT_MODE
+})
+@javax.annotation.Generated(
+    value = "org.openapitools.codegen.languages.JavaClientCodegen",
+    comments = "Generator version: 7.9.0")
+public class SearchDocumentKeyValueListPair {
```
newtork, line 31: (Major)
Looks like some classes do not have the `@Beta` annotation(?)

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/OrchestrationTest.java
MatKuhr: Note: this will be fixed by either AI/ai-sdk-java-backlog#110 or https://github.tools.sap/AI/llm-orchestration/pull/753

orchestration/src/main/java/com/sap/ai/sdk/orchestration/client/model/SearchDocumentKeyValueListPair.java
MatKuhr: This comes from not deleting the output directory before generating. To ensure we don't have any dead code, we should enable this. If it's not possible to use the OSS generator flag, we can use the maven clean plugin (refer to [this code](https://github.com/SAP/ai-sdk-java/commit/b7d7f6e8f8ba784aec0a6e0f5f9c72b8a1883ceb#diff-8d04401f1cc51365fe3e32f019cd720135ba920a1a7da7f19e9c9208478701fcL148)).

MatKuhr: I see we currently have some manually maintained classes in the package. Maybe we can move the manually maintained classes out such that we can safely delete and re-generate the package?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ConfigToRequestTransformer.java
```diff
@@ -42,14 +43,14 @@ static TemplatingModuleConfig toTemplateModuleConfig(
      * In this case, the request will fail, since the templating module will try to resolve the parameter.
      * To be fixed with https://github.tools.sap/AI/llm-orchestration/issues/662
      */
-    val messages = Option.of(template).map(TemplatingModuleConfig::getTemplate).getOrElse(List::of);
+    val messages = Option.of(template).map(t -> ((Template) t).getTemplate()).getOrElse(List::of);
```
MatKuhr, line 46: Not sure why the compiler doesn't complain about an unchecked cast here..

```suggestion
    val messages =
        Option.of(template)
            .filter(Template.class::isInstance)
            .map(Template.class::cast)
            .map(Template::getTemplate)
            .getOrElse(List::of);
```

orchestration/src/main/java/com/sap/ai/sdk/orchestration/client/model/SearchDocumentKeyValueListPair.java
newtork: > Maybe we can move the manually maintained classes out

Or further leverage the `mixIn` logic to override annotations.

orchestration/pom.xml
newtork: We cannot remove the annotation with just using configuration properties. We could provide a custom mustache file. But at this point, we can stop the additional effort. It's considered a temporary solution anyway.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/client/model/SearchDocumentKeyValueListPair.java
newtork: Done.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ConfigToRequestTransformer.java
newtork: > why

I assume, it's because the interface has only one known implementing class.

newtork (approved):

PR #157, opened 2024-11-13
Added Response Convenience

Jonas-Isr:
## Context

AI/ai-sdk-java-backlog#110.

As an orchestration user, I want to have a convenient way of accessing the LLM result and further related information.

### Feature scope:
 
- [x] Added ``OrchestrationResponse`` extends ``CompletionPostResponse``

## Definition of Done

- [x] Documentation updated
- [ ] ~Release notes updated~

### DIFF
--- a/orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationResponse.java
+++ b/orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationResponse.java
@@ -1,0 +1,26 @@
+package com.sap.ai.sdk.orchestration;
+
+import com.sap.ai.sdk.orchestration.client.model.CompletionPostResponse;
+import javax.annotation.Nonnull;
+
+/** Orchestration chat completion output. */
+public class OrchestrationResponse extends CompletionPostResponse {
+  /**
+   * Get the message content from the output.
+   *
+   * <p>Note: If there are multiple choices only the first one is returned
+   *
+   * @return the message content or empty string.
+   * @throws OrchestrationClientException if the content filter filtered the output.
+   */
+  @Nonnull
+  public String getContent() throws OrchestrationClientException {
+    if (getOrchestrationResult().getChoices().isEmpty()) {
+      return "";
+    }
+    if ("content_filter".equals(getOrchestrationResult().getChoices().get(0).getFinishReason())) {
+      throw new OrchestrationClientException("Content filter filtered the output.");
+    }
+    return getOrchestrationResult().getChoices().get(0).getMessage().getContent();
+  }
+}
--- a/pom.xml
+++ b/pom.xml
@@ -74,7 +74,7 @@
     <enforcer.skipBanGeneratedModulesReference>false</enforcer.skipBanGeneratedModulesReference>
     <!-- Test coverage -->
     <coverage.instruction>74%</coverage.instruction>
-    <coverage.branch>62%</coverage.branch>
+    <coverage.branch>60%</coverage.branch>
     <coverage.complexity>67%</coverage.complexity>
     <coverage.line>75%</coverage.line>
     <coverage.method>80%</coverage.method>
--- a/docs/guides/ORCHESTRATION_CHAT_COMPLETION.md
+++ b/docs/guides/ORCHESTRATION_CHAT_COMPLETION.md
@@ -91,8 +91,7 @@ var prompt = new OrchestrationPrompt("Hello world! Why is this phrase so famous?
 
 var result = client.chatCompletion(prompt, config);
 
-String messageResult =
-        result.getOrchestrationResult().getChoices().get(0).getMessage().getContent();
+String messageResult = result.getContent();
 ```
 
 In this example, the Orchestration service generates a response to the user message "Hello world! Why is this phrase so famous?".
--- a/orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
+++ b/orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
@@ -9,7 +9,6 @@ import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import com.sap.ai.sdk.core.AiCoreDeployment;
 import com.sap.ai.sdk.core.AiCoreService;
 import com.sap.ai.sdk.orchestration.client.model.CompletionPostRequest;
-import com.sap.ai.sdk.orchestration.client.model.CompletionPostResponse;
 import com.sap.ai.sdk.orchestration.client.model.ModuleConfigs;
 import com.sap.ai.sdk.orchestration.client.model.OrchestrationConfig;
 import com.sap.cloud.sdk.cloudplatform.connectivity.ApacheHttpClient5Accessor;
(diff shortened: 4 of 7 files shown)


orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationResponse.java
```diff
@@ -0,0 +1,26 @@
...
+   * Get the message content from the output.
+   *
+   * <p>Note: If there are multiple choices only the first one is returned
+   *
+   * @return the message content or empty string.
+   * @throws OrchestrationClientException if the content filter filtered the output.
+   */
+  @Nonnull
+  public String getContent() throws OrchestrationClientException {
+    if (getOrchestrationResult().getChoices().isEmpty()) {
```
MatKuhr, line 18: Can this ever happen? How is it handled in our OpenAI implementation ?

pom.xml
```diff
@@ -74,7 +74,7 @@
     <enforcer.skipBanGeneratedModulesReference>false</enforcer.skipBanGeneratedModulesReference>
     <!-- Test coverage -->
     <coverage.instruction>74%</coverage.instruction>
-    <coverage.branch>62%</coverage.branch>
+    <coverage.branch>60%</coverage.branch>
```
MatKuhr, line 77: We should probably instead improve our tests to cover the missing branches. I guess in this case this is for example the case that no choices are returned

MatKuhr (commented):
What about the other aspects like token usage?

Otherwise lgtm. While I would prefer a bit stronger separation to our generated code, I can't think of a specific scenario where the inheritance will break sonerhing. The main downside of the inheritance IMO is that the most used convenience methods get buried among the generated getters and setters. But currently there are not too many of those on the root response object, so should be fine

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationResponse.java
```diff
@@ -0,0 +1,26 @@
+package com.sap.ai.sdk.orchestration;
+
+import com.sap.ai.sdk.orchestration.client.model.CompletionPostResponse;
+import javax.annotation.Nonnull;
+
+/** Orchestration chat completion output. */
+public class OrchestrationResponse extends CompletionPostResponse {
```
newtork, line 7: As discussed, a solution without inheritance would be preferred.

pom.xml
Jonas-Isr: We did not manage to come up with a prompt to ChatGPT that would hit the output filter without already being stopped by the input filter. Thus, we were not able to test the (remaining) branch.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationResponse.java
Jonas-Isr: Removed (also in the Open AI implementation).

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationResponse.java
```diff
@@ -0,0 +1,33 @@
+package com.sap.ai.sdk.orchestration;
+
+import com.sap.ai.sdk.orchestration.client.model.CompletionPostResponse;
+import com.sap.ai.sdk.orchestration.client.model.LLMModuleResultSynchronous;
+import javax.annotation.Nonnull;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+/** Orchestration chat completion output. */
+@RequiredArgsConstructor
```
MatKuhr, line 10:
```suggestion
@RequiredArgsConstructor(access = PACKAGE)
```

Can be made package private, right?

MatKuhr: Also, add equals, hash code and tostring 

Maybe just use @value or consider a record

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/model/OpenAiChatCompletionOutput.java
```diff
@@ -40,9 +40,6 @@ public class OpenAiChatCompletionOutput extends OpenAiCompletionOutput
    */
   @Nonnull
   public String getContent() throws OpenAiClientException {
-    if (getChoices().isEmpty()) {
-      return "";
-    }
```
newtork, line 45: (Question)

What is the reason to remove this check? Would you rather risk a `NullPointerException´?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationResponse.java
```diff
@@ -0,0 +1,35 @@
...
+public class OrchestrationResponse {
+  CompletionPostResponse data;
+
+  /**
+   * Get the message content from the output.
+   *
+   * <p>Note: If there are multiple choices only the first one is returned
+   *
+   * @return the message content or empty string.
+   * @throws OrchestrationClientException if the content filter filtered the output.
```
newtork, line 23: (Major/Comment)

In case I'm not missing something, and we're ignoring the risk, let's at least document that(?)

```suggestion
   * @throws OrchestrationClientException if the content filter filtered the output.
   * @throws NullPointerException if the payload is incomplete.
```

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/model/OpenAiChatCompletionOutput.java
newtork: > You cannot have 0 choices per OpenAI docs.

I understand, but for me this would rather justify an exception than ignoring it :)

Not a strong opinion.

Jonas-Isr: Charles and I decided on reading this and adding tests.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationResponse.java
Jonas-Isr: See comment above.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationResponse.java
```diff
@@ -0,0 +1,35 @@
...
+import com.sap.ai.sdk.orchestration.client.model.LLMModuleResultSynchronous;
+import javax.annotation.Nonnull;
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+
+/** Orchestration chat completion output. */
+@Value
+@RequiredArgsConstructor(access = PACKAGE)
+public class OrchestrationResponse {
+  CompletionPostResponse data;
```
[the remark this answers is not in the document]
Jonas-Isr: Renamed the filed to `originalResponse`.

MatKuhr (dismissed):
LGTM 👍🏻

PR #168, opened 2024-11-18
fix: [DevOps] Release Script

MatKuhr:
## Context

AI/ai-sdk-java-backlog#115.

### DIFF
--- a/.github/workflows/perform-release.yml
+++ b/.github/workflows/perform-release.yml
@@ -114,7 +114,7 @@ jobs:
       - name: "Deploy"
         run: |
           MVN_ARGS="${{ env.MVN_CLI_ARGS }} -Drelease -s settings.xml"
-          mvn org.sonatype.central:central-publishing-maven-plugin:publish $MVN_ARGS
+          mvn deploy $MVN_ARGS
         env:
           MAVEN_GPG_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/perform-release.yml
```diff
@@ -113,8 +113,8 @@ jobs:
 
       - name: "Deploy"
         run: |
-          MVN_ARGS="${{ env.MVN_CLI_ARGS }} deploy -Drelease -s settings.xml"
-          mvn $MVN_ARGS
+          MVN_ARGS="${{ env.MVN_CLI_ARGS }} -Drelease -s settings.xml"
+          mvn deploy $MVN_ARGS
```
MatKuhr, line 117: I tried using `mvn org.sonatype.central:central-publishing-maven-plugin:publish`, but that doesn't work on its own

pom.xml
```diff
@@ -724,6 +724,28 @@ https://gitbox.apache.org/repos/asf?p=maven-pmd-plugin.git;a=blob_plain;f=src/ma
           </plugins>
         </pluginManagement>
         <plugins>
+          <!-- The release artifacts don't contain our custom config files for these plugins -->
```
MatKuhr, line 727: in theory, many other plugins could be disabled for this, but it's annoying to maintain, so we only disable those that require custom files not part of the release tarball

.github/workflows/perform-release.yml
```diff
@@ -88,9 +88,6 @@ jobs:
         with:
           distribution: "sapmachine"
           java-version: ${{ env.JAVA_VERSION }}
-          server-id: ossrh
```
MatKuhr, line 91: I'm assuming these are leftovers because we are no longer pushing to OSSRH

CharlesDuboisSAP (approved):

PR #177, opened 2024-11-21
feat: [Orchestration] Configuration as JSON String

MatKuhr:
## Context

This PR adds a feature to directly use JSON as downloaded from AI Launchpad.

## Definition of Done

- [x] Documentation updated
- [x] ~Release notes updated~


orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
```diff
@@ -144,6 +147,54 @@ public CompletionPostResponse executeRequest(@Nonnull final CompletionPostReques
...
+   *
+   * @param prompt The input parameters and optionally message history to use for prompt execution.
+   * @param moduleConfig The module configuration in JSON format.
+   * @return The completion response.
+   * @throws OrchestrationClientException If the request fails.
+   */
+  @Beta
+  @Nonnull
+  public OrchestrationChatResponse executeRequestFromJsonModuleConfig(
+      @Nonnull final OrchestrationPrompt prompt, @Nonnull final String moduleConfig)
```
CharlesDuboisSAP, line 166:
```suggestion
  public OrchestrationChatResponse executeJsonRequest(
      @Nonnull final OrchestrationPrompt prompt, @Nonnull final String json)
```
It's not called a module config in AI Launchpad right? It's only called "JSON"

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
```diff
@@ -144,6 +147,54 @@ public CompletionPostResponse executeRequest(@Nonnull final CompletionPostReques
...
+      throws OrchestrationClientException {
+    if (!prompt.getMessages().isEmpty()) {
+      throw new IllegalArgumentException(
+          "Prompt must not contain any messages when using a JSON module configuration, as the template is already defined in the JSON.");
+    }
```
CharlesDuboisSAP, line 171: An OrchestrationPrompt  is not the right object then. Why not accept a map of input params?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
```diff
@@ -144,6 +147,54 @@ public CompletionPostResponse executeRequest(@Nonnull final CompletionPostReques
...
+   * <p><b>NOTE:</b> This method does not support streaming.
+   *
+   * @param prompt The input parameters and optionally message history to use for prompt execution.
+   * @param moduleConfig The module configuration in JSON format.
+   * @return The completion response.
+   * @throws OrchestrationClientException If the request fails.
+   */
+  @Beta
+  @Nonnull
+  public OrchestrationChatResponse executeRequestFromJsonModuleConfig(
```
newtork, line 165: (Comment)
I don't know, it feels like this is method naming gore 😁

docs/guides/ORCHESTRATION_CHAT_COMPLETION.md
```diff
@@ -208,3 +208,19 @@ OrchestrationAiModel customGPT4O =
                 "presence_penalty", 0))
         .withModelVersion("2024-05-13");
 ```
+
+### Using a Configuration from AI Launchpad
```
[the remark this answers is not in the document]
MatKuhr: How would you show that, maybe with a screenshot?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
MatKuhr: Because that is missing the (optional) messages history. 

> OrchestrationPrompt is not the right object

I'd argue it is. We will have further cases where passing messages is not allowed, e.g. when using TemplateRef. If we want we could create another Prompt object for that to make this distinction, but so far I figured it's not worth it.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
MatKuhr: I didn't name it that way because we don't expect the full request JSON. In Launchpad it's called `Orchestration workflow configuration`

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
MatKuhr: felt the same, I reduced it a little bit now

orchestration/src/test/java/com/sap/ai/sdk/orchestration/OrchestrationUnitTest.java
```diff
@@ -393,4 +393,54 @@ void testEmptyChoicesResponse() {
 
     assertThat(result.getContent()).isEmpty();
   }
+
+  @Test
+  void testExecuteRequestFromJson() {
+    stubFor(post(anyUrl()).willReturn(okJson("{}")));
+
+    prompt = new OrchestrationPrompt(Map.of());
```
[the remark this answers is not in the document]
jjtang1985: Please have a look at the following json, that I copied from the Launchpad.

```json
{
    "module_configurations": {
        "llm_module_config": {
            "model_name": "mistralai--mistral-large-instruct",
            "model_params": {},
            "model_version": "2407"
        },
        "templating_module_config": {
            "template": [
                {
                    "role": "user",
                    "content": "How are you?"
                }
            ],
            "defaults": {}
        }
    }
}
```

I would assume, I've got everything from the JSON and should be able to run it with a function like
```
myFunction(json);
```
The second param is only needed (so it should be optional), when I want to pass a `inputParam` or providing message history like the following screenshot:

Screenshot 2024-11-21 at 20 55 47

docs/guides/ORCHESTRATION_CHAT_COMPLETION.md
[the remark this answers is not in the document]
jjtang1985: Currently, when reading this, I would only copy/paste the JSON.
In addition to that, you can also use the downloaded json file.
To do so, you need to read the json file.
We can add either a sentence for that or a code example.

docs/guides/ORCHESTRATION_CHAT_COMPLETION.md
```diff
@@ -208,3 +208,19 @@ OrchestrationAiModel customGPT4O =
...
+In case you have created a configuration in AI Launchpad, you can copy or download the configuration as JSON and use it directly in your code:
+
+```java
+var configJson = """
+    ... paste your configuration JSON in here ...
+    """;
+
+var prompt = new OrchestrationPrompt(Map.of("your-input-parameter", "your-param-value"));
+
+new OrchestrationClient().executeRequestFromJsonModuleConfig(prompt, configJson);
```
newtork, line 223: Your suggested code change results in the following two methods


```java
OrchestrationClient client = new OrchestrationClient();

// execute completion request (exist)
CompletionPostRequest requestDto;
CompletionPostResponse responseDto = client
  .executeRequest(
    requestDto);

// execute prompt (exist)
OrchestrationPrompt promptConvenience;
OrchestrationModuleConfig configConvenience;
OrchestrationChatResponse responseConvenience = client
  .executeRequest(
    promptConvenience,
    configConvenience);

// execute prompt (new)
OrchestrationPrompt promptConvenience;
String configJson;
OrchestrationChatResponse responseConvenience = client
  .executeRequestFromJsonModuleConfig(
    promptConvenience,
    configJson);
```

(Question)

Is it not possible to derive reuse existing method and allow for `OrchestrationModuleConfig` to be created from `String configJson`? That would look a little cleaner, maybe(?)

MatKuhr: I also thought about this approach. But unfortunately not, because the JSON string is expected to contain things not yet supported by `OrchestrationModuleConfig` or it's individual config objects, i.e. throughout the entire JSON. That includes cases of `oneOf`, e.g. TemplateRef.

orchestration/src/test/java/com/sap/ai/sdk/orchestration/OrchestrationUnitTest.java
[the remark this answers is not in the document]
MatKuhr: Ah true, Launchpad doesn't force you to use any variables. Okay, will update 👍🏻

docs/guides/ORCHESTRATION_CHAT_COMPLETION.md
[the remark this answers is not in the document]
MatKuhr: will add a comment 👍🏻

orchestration/src/test/java/com/sap/ai/sdk/orchestration/OrchestrationUnitTest.java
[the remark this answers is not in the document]
MatKuhr: After thinking about this again, I actually think the use case for this is not very strong. Because if there are no input params or messages history, prompt execution will always yield the same or very similar results (depending on the temperature setting).

The same is true when using template references. If we really want to support such usage, we could consider allowing `new OrchestrationPrompt()` for these cases. But making the `prompt` itself optional would mean we add 2-3 method overloads for non-productive cases..

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
```diff
@@ -132,26 +134,75 @@ public OrchestrationChatResponse chatCompletion(
...
+
+    final ObjectNode requestJson = JACKSON.valueToTree(request);
+    final JsonNode moduleConfigJson;
+    try {
+      moduleConfigJson = JACKSON.readTree(moduleConfig);
+    } catch (JsonProcessingException e) {
+      throw new IllegalArgumentException(
+          "The provided module configuration is not valid JSON: " + moduleConfig, e);
+    }
+    requestJson.set("orchestration_config", moduleConfigJson);
```
newtork, line 184: (Major/Question)

Wouldn't it make more sense applying the module config JSON to client instead of per request?
Seems like the only place where `orchestration_config` is provided, could be an optional parameter that is passed along the convenience API(?)

MatKuhr: Hm not exactly sure what you mean.. Do you mean setting the config on the `OrchestrationClient` object instead? Or something else?

newtork: Yes.
Boiling this down to the question: Is `orchestration_config` to be expected **request specific** or **client specific**?

MatKuhr: Essentially same as the `OrchestrationConfig` object: It depends how much, if any, config changes per request. But the tendency is it is tied to the client, i.e. for typical use cases the majority of config will stay the same across multiple requests.

However, we decided to not allow config per client to make the API simpler. So the same strategy is applied for this method too. If we want to, we can still change this in the future though.

newtork: To me, from a user perspective, it would feel weird to pass a `String` object along (over multiple controller classes potentially) just to pass to my request invocations.

But it's difficult to argue about client / request specific data handling, when I haven't yet seen a proper use-case "in the wild". Maybe we will discuss this again, after having users first interact with that.

newtork (approved):

PR #178, opened 2024-11-21
fix: AiCoreService NoClassDefFound Error

MatKuhr:
## Context

AI/ai-sdk-java-backlog#120.

This PR fixes an issue where malformed JSON in a `.env` file or `AICORE_SERVICE_KEY` env var would lead to a `NoClassDefFound`.

Contains a small refactoring to allow proper testability, done via #181 

## Definition of Done

- [x] ~Documentation updated~
- [x] ~Release notes updated~


newtork (commented):
I don't feel comfortable, merging this without the changes of #181 

The latter I would like to see released.

core/src/test/java/com/sap/ai/sdk/core/AiCoreServiceKeyAccessorTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+    var accessor = new AiCoreServiceKeyAccessor(dotenv);
+    assertThat(accessor.getServiceBindings()).isNotEmpty().hasSize(1);
+    assertThat(accessor.getServiceBindings().get(0).getCredentials())
+        .containsEntry("clientid", "cid")
+        .containsEntry("clientsecret", "csecret")
+        .containsEntry("url", "https://authenticationsap")
+        .containsEntry("serviceurls", Map.of("AI_API_URL", "https://api.ai.sap"));
+
+    verify(dotenv, times(1)).load();
```
newtork, line 28: (Minor)

Thanks for adding the test. While this makes sure we're not running I/O heavy task multiple times, the logic still runs `ServiceBinding createServiceBinding(@Nonnull final String serviceKey)` repeatedly. for every AI Core interaction.

Is this accepted / expected?

MatKuhr: Yeah, also not super happy with that, but considering this is recommended only for non-prod, at least acceptable. I think we can later on add a cache, but then probably rather on the destination object, as caching that skips further object instatiation

PR #180, opened 2024-11-21
[Orchestration Convenience] Typed model parameters

newtork:
Would be helpful for demo purposes and general usability.

```diff
+ import static com.sap.ai.sdk.orchestration.OrchestrationAiModel.Parameter.*;

OrchestrationAiModel customGPT4O =
    OrchestrationAiModel.GPT_4O
-       .withParams(
-           Map.of(
-               "max_tokens", 50,
-               "temperature", 0.1,
-               "frequency_penalty", 0,
-               "presence_penalty", 0))
+       .withParam(MAX_TOKENS, 50)
+       .withParam(TEMPERATURE, 0.1)
+       .withParam(FREQUENCY_PENALTY, 0)
+       .withParam(PRESENCE_PENALTY, 0)
        .withVersion("2024-05-13");
```

image


orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModelParameters.java
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * Get the parameters.
+   *
+   * @return the parameters.
+   */
+  @Nonnull
+  Map<String, Object> getParams();
+
+  /** GPT model parameters. */
+  interface GPT extends OrchestrationAiModelParameters {
```
jjtang1985, line 18: [q] Are these GPT specific (model) parameters or Azure specific (platform/Rest API)?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModelParameters.java
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * Create a new builder.
+     *
+     * @return the builder.
+     */
+    @Nonnull
+    static GPT.Builder0 params() {
```
MatKuhr, line 25: Considering all parameters are optional, why is this a builder?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModelParameters.java
```diff
@@ -0,0 +1,86 @@
...
+      return maxTokens ->
+          temperature ->
+              frequencyPenalty ->
+                  presencePenalty ->
+                      () ->
+                          Map.of(
+                              "max_tokens", maxTokens,
```
MatKuhr, line 32: why this specific sub-selection of (Azure) OpenAI parameters?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModel.java
```diff
@@ -114,4 +123,28 @@ public class OrchestrationAiModel {
   LLMModuleConfig createConfig() {
     return new LLMModuleConfig().modelName(name).modelParams(params).modelVersion(version);
   }
+
+  /**
+   * Subclass to allow for parameterized models.
+   *
+   * @param <T> The type of parameters for this model.
+   */
+  public static final class Parameterized<T extends OrchestrationAiModelParameters>
```
MatKuhr, line 132: why do we need this class?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModelParameters.java
newtork: > Considering all parameters are optional

I was under the assumption, all parameters are required.

MatKuhr: no, those are just a somewhat random selection to give examples

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModelParameters.java
newtork: After talking to Orchestration team, I was made aware, these parameters are expected to work with all LLM providers (today). They are mapping the above OpenAI related parameter to all supported providers.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModelParameters.java
newtork: Only after asking Orchestration team, I realized they are supporting the parameters for non-OpenAI LLM providers as well. From the spec that wasn't apparent. As a result, we can change the solution to be model-agnostic.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModel.java
newtork: Assuming the [LLM providers expect individual sets of parameters](https://docs.litellm.ai/docs/completion/input#translated-openai-params) the additional `withParams( T )` would require a generic parameter type argument. As you can see VertexAI would not understand all your OpenAI arguments. To model this distinction a specific type argument (i.e. type hinting) is advantageous.

newtork (commented):
As discussed in todays (after) daily..
* **We want** to offer API (via fields, methods or classes) for the available parameter keys.
* **We can't** reasonably maintain the full mapping of LLM provider / model parameters.
* **We can** carefully(?) assume a very small subset of keys to be available for all LLM providers.

How to translate that to code, needs to be aligned still.

jjtang1985 (dismissed):
The API looks good!
Very helpful.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModel.java
```diff
@@ -114,4 +117,58 @@ public class OrchestrationAiModel {
...
+    /** The maximum number of tokens for completion */
+    MAX_COMPLETION_TOKENS("max_completion_tokens"),
+    /** The probability mass to be considered . */
+    TOP_P("top_p"),
+    /** The toggle to enable partial message delta. */
+    STREAM("stream"),
+    /** The options for streaming response. */
+    STREAM_OPTIONS("stream_options"),
+    /** The tokens where the API will stop generating further tokens. */
+    STOP("stop"),
```
MatKuhr, line 168: Isn't this the same as `max_tokens` then?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModel.java
```diff
@@ -114,4 +117,58 @@ public class OrchestrationAiModel {
...
+  public OrchestrationAiModel withParam(
+      @Nonnull final Parameter param, @Nullable final Object value) {
+    return withParam(param.value, value);
+  }
+
+  /** Parameter key for a model. */
+  @RequiredArgsConstructor
+  public enum Parameter {
+    /** The maximum number of tokens to generate. */
+    MAX_TOKENS("max_tokens"),
```
MatKuhr, line 152: Do you think it would be feasible to also include the datatype class for these properties?

e.g. something like:
```suggestion
    MAX_TOKENS("max_tokens", Integer.class),
```

and use a generic in the `withParam` to type-check the value being passed on the API level? Or maybe this is overkill for this API?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModel.java
newtork: If I understood correctly, it refers to a stop token / literal / string, on which it will stop generating - or on which it signals generation stop.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModel.java
newtork: I like the idea, changing the solution was easy enough.

MatKuhr (approved):
LGTM

PR #181, opened 2024-11-22
chore: [Core] Refactor Service Key Loading and Testing

MatKuhr:
## Context

Follow up to #178 

### Feature scope:
 
- [x] Test loading of `.env` files
- [x] Refactor service key loading code to be unit testable

## Definition of Done

- [x] ~Documentation updated~
- [x] ~Release notes updated~


core/src/main/java/com/sap/ai/sdk/core/AiCoreCredentialsInvalidException.java
```diff
@@ -0,0 +1,7 @@
+package com.sap.ai.sdk.core;
+
+import lombok.experimental.StandardException;
+
+/** Exception thrown when the JSON AI Core service key is invalid. */
+@StandardException
+class AiCoreCredentialsInvalidException extends RuntimeException {}
```
MatKuhr, line 7: ([moved from static inner class](https://github.com/SAP/ai-sdk-java/pull/181/files#diff-4da9ee11275233c1ba8615a9f7636b44cc2247b32afb28efa426c143cf545dadL103)) 

The static inner class lead to the name printed out being `com.sap.ai.sdk.core.DestinationResolver$AiCoreCredentialsInvalidException` which isn't great...

core/src/main/java/com/sap/ai/sdk/core/AiCoreDeployment.java
```diff
@@ -30,7 +30,8 @@ public AiCoreDeployment withResourceGroup(@Nonnull final String resourceGroup) {
 
   @Nonnull
   @Override
-  public Destination destination() throws DestinationAccessException, DestinationNotFoundException {
+  public HttpDestination destination()
```
MatKuhr, line 33: Relieves users from unnecessary `.asHttp()` calls

core/src/main/java/com/sap/ai/sdk/core/DestinationResolver.java
```diff
@@ -1,119 +1,57 @@
...
-  static HttpDestination getDestination(@Nullable final String serviceKey) {
-    final Predicate<Object> aiCore = Optional.of(ServiceIdentifier.AI_CORE)::equals;
-    val serviceBindings = accessor.getServiceBindings();
-    val aiCoreBinding = tryFind(serviceBindings, b -> aiCore.test(b.getServiceIdentifier()));
-
-    val serviceKeyPresent = serviceKey != null;
-    if (!aiCoreBinding.isPresent() && serviceKeyPresent) {
-      addServiceBinding(serviceKey);
-    }
+  HttpDestination getDestination() {
```
MatKuhr, line 34: no longer static for better testability

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
```diff
@@ -31,19 +27,21 @@
...
+  static final String AI_RESOURCE_GROUP = "URL.headers.AI-Resource-Group";
 
   private static final DeploymentCache DEPLOYMENT_CACHE = new DeploymentCache();
 
-  private static final String AI_RESOURCE_GROUP = "URL.headers.AI-Resource-Group";
+  @Nonnull private final DestinationResolver destinationResolver;
 
-  /** loads the .env file from the root of the project */
-  private static final Dotenv DOTENV = Dotenv.configure().ignoreIfMissing().load();
+  @Nonnull private Function<AiCoreService, HttpDestination> baseDestinationHandler;
```
newtork, line 39: (Question)

Unrelated to your PR I'm wondering why this field is not final.

core/src/main/java/com/sap/ai/sdk/core/DestinationResolver.java
```diff
@@ -1,119 +1,57 @@
...
+  @Nonnull private final ServiceBindingAccessor accessor;
 
-  @Getter(AccessLevel.PROTECTED)
-  @Nonnull
-  private static ServiceBindingAccessor accessor = DefaultServiceBindingAccessor.getInstance();
+  DestinationResolver() {
+    this(
+        new ServiceBindingMerger(
+            List.of(DefaultServiceBindingAccessor.getInstance(), new AiCoreServiceKeyAccessor()),
+            ServiceBindingMerger.KEEP_EVERYTHING));
```
newtork, line 30: (Major/Question)

I see no cache in place, does that mean the service bindings are resolved again with every AI Core interaction?

core/src/main/java/com/sap/ai/sdk/core/AiCoreCredentialsInvalidException.java
newtork: I would even make it public at this point.

core/src/main/java/com/sap/ai/sdk/core/AiCoreDeployment.java
MatKuhr: Users might want to access the URL and headers e.g. to use an HTTP client of their choice for requests. But that is rather an edge case, I was mainly concerned with java generics, where the difference between interface and implementation can become annoying. And I'd hope we don't do destination API changes in v6 😄 

But this is really just a minor change, using `Destination` is also fine IMO. Just felt a little more intuitive to me..

core/src/main/java/com/sap/ai/sdk/core/AiCoreCredentialsInvalidException.java
MatKuhr: Also thought about that, but so far didn't do it because I figured users should never catch this exception explicitly, it's just there to be more descriptive. So no benefit really in making it public, we could also remove it entirely and fully rely on the exception message to do the job. No strong opinion

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
MatKuhr: Was wondering that too and planning to create another separate PR after this one with changes

core/src/main/java/com/sap/ai/sdk/core/DestinationResolver.java
MatKuhr: The default accessor will have a cache for the "normal" bindings. For the service key accessor, I now added a test that verifies files are only loaded once.

PR #191, opened 2024-11-26
fix: Update Response Masking Format

MatKuhr:
## Context

This PR adjust the tests to reflect a change of in the orchestration response regarding masking.

Note: we assert on the response format to ensure the examples are working as intended, even though the response format is not part of the spec and might change in the future


CharlesDuboisSAP (approved):
A beautiful improvement

PR #195, opened 2024-11-27
Chat message convenience

rpanackal:
## Context

AI/ai-sdk-java-backlog#101.

Introduce convenient API for creating chat messages.

### Feature scope:
 
- [x] Add assistant, user and system chat message classes

## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~

### DIFF
--- a/docs/adrs/005-release-strategy.md
+++ b/docs/adrs/005-release-strategy.md
@@ -1,0 +1,102 @@
+# Development Lifecycle
+
+Status: In-Progress.
+
+## Background
+
+The AI SDKs largely depend on the AI Core service, which currently is released roughly every 2 weeks.
+The release process itself (from the first integration landscape to the last production landscape) also takes around 2 weeks.
+
+This raises the question when to release new versions of the SDKs.
+
+## Development
+
+For any AI Core functionality not yet publicly released, the following options can be considered for the AI SDKs:
+
+- Allow no code
+- Allow test code
+- Allow (released) internal code
+- Allow (released) public API
+
+Note that for JavaScript and Python, code can be marked as internal and thus be considered _non-public_.
+For Java, the options are more limited: Manually written code can be marked package-private, but generated code can not.
+Since most new features will rely on generated code, the options for Java are limited.
+
+## Testing
+
+Testing can be performed against multiple landscapes and can vary based on the type of tests.
+
+- Unit tests may use test data from any landscape (including development landscapes) manually tested with e.g. Bruno.
+- E2E tests may use canary or production landscapes.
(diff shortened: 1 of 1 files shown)


orchestration/src/main/java/com/sap/ai/sdk/orchestration/AssistantMessage.java
```diff
@@ -0,0 +1,11 @@
+package com.sap.ai.sdk.orchestration;
+
+import com.sap.ai.sdk.orchestration.client.model.ChatMessage;
+import javax.annotation.Nonnull;
+
+public class AssistantMessage extends ChatMessage {
```
newtork, line 6: (Major)

I think we aligned on not-using inheritance between convenience- and low-level API.

For example see `OrchestrationAiModel`.

Please correct me if my observation / conclusion is wrong, @MatKuhr

MatKuhr: Yes, it's safer to clearly separate the two APIs

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AssistantMessage.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.sap.ai.sdk.orchestration.client.model.ChatMessage;
+import javax.annotation.Nonnull;
+
+/**
+ * Represents a chat message as 'assistant' to the orchestration service.
+ *
+ * @param content
+ */
+public record AssistantMessage(@Nonnull String content) implements Message {
```
MatKuhr, line 11: I think I'd make this a class, because the type of content will change in the future. Records are not as easily extensible and make the getters public, which is an issue if content needs to be a complex type

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationPrompt.java
```diff
@@ -32,7 +32,7 @@ public class OrchestrationPrompt {
    * @param message A user message.
    */
   public OrchestrationPrompt(@Nonnull final String message) {
-    messages.add(new ChatMessage().role("user").content(message));
+    messages.add(new UserMessage(message).toChatMessage());
```
MatKuhr, line 35: Instead of changing every usage here, it's probably easier to make the private fields of this object to accept the convenience class, and only convert later once. Then the two apis are also better separated

MatKuhr (commented):
Docs also need updates

PR #201, opened 2024-11-28
Refactor AiCoreService class

MatKuhr:
## Context

Follow up of #79 

This PR simplifies the core classes and reduces public API.

This comes at the cost of no longer supporting a flow where e.g. a scenario for a deployment lookup is given first, and a resource group given afterwards. Both now always have to be given together. This way, we avoid complicated state changes via side effects and potentially unexpected behavior.

### Feature scope:
 
- [x] Refactor `AiCoreService` class
  - [x] Updated base destination customization
  - [x] Reduced public API
  - [x] Updated generated code, requires SAP/cloud-sdk-java#652 
  - [x] `AiCoreService` is now immutable
  - [x] Updated the ADR 004
- [x] Fix race condition in `DeploymentResolver` 
- [x] More comprehensive tests for `DeploymentResolver`

## Definition of Done

- [x] Documentation updated
- [x] ~Release notes updated~


core/src/main/java/com/sap/ai/sdk/core/DeploymentResolver.java
```diff
@@ -19,99 +19,104 @@
...
+
+  @Nonnull
+  Optional<String> resolveDeployment(
+      @Nonnull final String resourceGroup, @Nonnull final Predicate<AiDeployment> predicate) {
+    var deployment = getCachedDeployment(resourceGroup, predicate);
+    if (deployment.isPresent()) {
+      return deployment;
+    }
+    // double-checked locking to allow for safe cache resets under parallel access
+    synchronized (cache) {
```
newtork, line 106: (Comment)

Not a fan of the double lock code. Can't this be solved with convenience API somewhere from the (vanilla) java collection API?

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
```diff
@@ -225,6 +182,6 @@ protected ApiClient buildApiClient(@Nonnull final Destination destination) {
    * @param resourceGroup the resource group of the deleted deployment, usually "default".
    */
   public void reloadCachedDeployments(@Nonnull final String resourceGroup) {
-    DEPLOYMENT_CACHE.resetCache(this, resourceGroup);
+    deploymentResolver.reloadDeployments(resourceGroup);
```
newtork, line 185: (Minor)

I understand your PR didn't introduce it. But I'm not sure whether we should pollute this class with cache mechanics :)

Not sure whether double locking is necessary here.

Can we annotate it `@Beta`?

core/src/main/java/com/sap/ai/sdk/core/client/ExecutableApi.java
```diff
@@ -33,16 +32,16 @@ public class ExecutableApi extends AbstractOpenApiService {
 
   /** Instantiates this API class to invoke operations on the AI Core */
   public ExecutableApi() {
-    super(new AiCoreService().client());
+    super(new AiCoreService().getApiClient());
```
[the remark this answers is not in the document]
MatKuhr: We have to stop with any generator change until the SDK is released. So in this case with the oneOf change..

core/src/main/java/com/sap/ai/sdk/core/DeploymentResolver.java
MatKuhr: Yeah, also don't like implementing this manually. But not sure. We can't directly use computeIfAbsent, because that would only compute when the resource group isn't found in the cache. I don't know of any other convenience / tooling around this that standardizes this..

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
MatKuhr: Agreed. In general, currently `AiCoreService` is just re-exposing the API defined by `DeploymentResolver`. I was contemplating to just make `DeploymentResolver` public, getting rid of this duplication. But that would come at the downside that the usage gets a bit more verbose:

```java
new AiCoreService().getDestinationForDeploymentByModel(..)
new AiCoreService().getDeploymentResolver().getDestinationForDeploymentByModel(..)
```

No strong opinion. We could also mark all the `getDestinationForDeployment` as beta to allow for future changes..

MatKuhr: (marked the cache reload as beta now)

core/src/main/java/com/sap/ai/sdk/core/client/ExecutableApi.java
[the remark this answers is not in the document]
newtork: Okay, I thought `core` module / spec was unaffected of this.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
```diff
@@ -57,27 +59,32 @@ public class OrchestrationClient {
...
    * <pre>{@code
-   * new OrchestrationClient(new AiCoreService().forDeploymentByScenario("orchestration"));
+   * new OrchestrationClient(new AiCoreService().getDestinationForDeploymentByScenario("custom-rg", "orchestration"));
    * }</pre>
    *
-   * @param deployment The specific {@link AiCoreDeployment} to use.
+   * @param destination The specific {@link HttpDestination} to use.
+   * @see AiCoreService#getDestinationForDeploymentByScenario(String, String)
    */
-  public OrchestrationClient(@Nonnull final AiCoreDeployment deployment) {
```
newtork, line 79: (Question/Minor)

I think after the change you need to document how to interact with orchestration service with `resource-group <> "default"`.

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
```diff
@@ -25,123 +24,94 @@
...
+   *     case of invalid credentials.
+   * @throws DestinationNotFoundException If there was an issue creating the base destination, e.g.
+   *     in case of missing credentials.
+   * @see #withBaseDestination(HttpDestination)
    */
   @Nonnull
-  public AiCoreService withDestination(@Nonnull final HttpDestination destination) {
-    baseDestinationHandler = service -> destination;
-    return this;
+  public HttpDestination getBaseDestination()
```
newtork, line 85: (Question/Minor)

What's the use-case for this method being public for users?

MatKuhr: This is what we could recommend for anyone who want to execute requests against AI Core manually e.g. using our HTTP client. That way the base path + client type header are already set, and it's much easier, compared to using the `ServiceBindingDestinationLoader` logic directly.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
MatKuhr: The level of documentation doesn't really change before / after this PR I think..? But yes, should be e2e tested and documented as part of the [related ticket](https://github.tools.sap/AI/ai-sdk-java-backlog/issues/123), thus I would leave that out of this PR.

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
```diff
@@ -150,54 +120,29 @@ public AiCoreDeployment forDeploymentByModel(@Nonnull final AiModel model)
...
-      @Nonnull final Destination destination) {
-    val builder = DefaultHttpDestination.fromDestination(destination);
-    String uri = destination.get(DestinationProperty.URI).get();
-    if (!uri.endsWith("/")) {
-      uri = uri + "/";
-    }
-    builder.uri(uri + "v2/").property(AI_CLIENT_TYPE_KEY, AI_CLIENT_TYPE_VALUE);
-    return builder;
-  }
+  public ApiClient getApiClient() {
```
newtork, line 141: (Question)

Didn't we find a way to mitigate `ApiClient`?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
newtork: I just realized, you guys botched immutability for `withResourceGroup(...)` in #100.

I guess, that's on me for not adding unit tests for every immutability contract ._.

core/src/main/java/com/sap/ai/sdk/core/DestinationResolver.java
```diff
@@ -48,7 +55,29 @@ HttpDestination getDestination() {
...
+  static HttpDestination fromCustomBaseDestination(@Nonnull final HttpDestination destination) {
+    // for custom base destinations we only add the client type header, since users are allowed to
+    // pass a custom base path
+    return addClientTypeHeader(destination);
+  }
+
+  @Nonnull
+  private static HttpDestination setBasePath(@Nonnull final HttpDestination destination) {
+    return DefaultHttpDestination.fromDestination(destination)
+        .uri(destination.getUri().resolve(BASE_PATH))
```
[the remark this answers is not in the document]
MatKuhr: The base path is intentionally absolute and will override any path given in the service binding. At least for now, the binding should never contain any path, so any given path would likely be faulty.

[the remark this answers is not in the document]
newtork: > the binding should never contain any path

Feels like an undocumented assumption.
We don't know where the service binding is coming from.

core/src/main/java/com/sap/ai/sdk/core/DeploymentResolver.java
newtork: I see. Maybe one option to do this properly would be a real cache implementation, that has invalidation handlers - which would trigger when _predicate_ returned _empty_. But I understand that sounds like overkill for the current code line.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
newtork: > The level of documentation doesn't really change before / after this PR I think..?

To me this is surprising too... I assumed we had docs on that. We can discuss in Slack. I agree, it's not a requirement for your PR.

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
newtork: Okay, while reviewing I wasn't aware, you are promoting `HttpDestination` as potential adapter between the APIs

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
MatKuhr: Yes, I think in part that is the reason, but also we want to return something that is already pre-configured with the right Jackson so that it works OOTB

core/src/main/java/com/sap/ai/sdk/core/DestinationResolver.java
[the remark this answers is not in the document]
MatKuhr: As long as users use a service binding produced by AI Core, this guaranteed.

Both approaches, merging paths as well as removing paths, can lead to confusion. But I think removing leads to less confusion, because it doesn't depend on user input, i.e. wether or not users add a trailing slash or not.

So yes, agree resolve can be confusing, I just hope this solution produces the least confusion 😄

MatKuhr:
Added an alternative with a Builder: #211

core/src/main/java/com/sap/ai/sdk/core/AiCoreService.java
MatKuhr: Agreed, done ✅

newtork (approved):
lgtm

PR #219, opened 2024-12-03
change base path behaviour

MatKuhr:
## Context

Minor change after discussing with JS team, having the base path set automatically is the more frequent use case


newtork (approved):
It's a shame this path manipulation is so error prone in Java.

Maybe there are fewer pitfalls with `URL` instead of `URI` 🤔&lt;/cloud.sdk>

PR #222, opened 2024-12-03
E2E Different Resource Group

rpanackal:
## Context

AI/ai-sdk-java-backlog#123.

Include test to cover using OpenAI or Orchestration service with a different `resourceGroup` than `"default"`.

### Feature scope:
 
- [x] Add new endpoints in sample app for filtering completions by resource group
- [x] Fix bug in using custom destination from OpenAI client
  - [x] All calls to OpenAI service declares a default api version by default (unless explicitly set)

## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~


foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
```diff
@@ -74,7 +74,9 @@ void apiVersion() {
     stubFor(post(anyUrl()).willReturn(okJson("{}")));
     Try.of(() -> client.chatCompletion(new OpenAiChatCompletionParameters()));
 
-    verify(exactly(1), postRequestedFor(anyUrl()).withoutQueryParam("api-version"));
+    verify(
+        exactly(1),
+        postRequestedFor(anyUrl()).withQueryParam("api-version", equalTo("2024-02-01")));
```
newtork, line 79: (Explanation)

Without the `api-version` we experienced _resource not found_ error messages from AI Core.

sample-code/spring-app/pom.xml
```diff
@@ -157,6 +157,10 @@
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.sap.cloud.sdk.cloudplatform</groupId>
+      <artifactId>cloudplatform-connectivity</artifactId>
+    </dependency>
```
newtork, line 163: (Major)
Wait, why is this necessary?

newtork (approved):

sample-code/spring-app/pom.xml
rpanackal: Without the declaration, the maven dependency plugin with goal "analyze-only" reports "Used undeclared dependencies found"

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
MatKuhr: Correct, this was intended. The idea being that it is possible for users to prevent the api version from being set, e.g. for cases where the some model works without api version, or maybe it might be set automatically by long access; or if the key of that header ever changes.

That is why the javadoc of the method explicitly states to call withApiVersion afterwards 


But to be fair, the above are rather hypothetical scenarios and i think it may be worthwhile to change it in favor of the more convenient approach of setting this automatically

MatKuhr (approved):
Adding a small section in the docs for this would be useful, otherwise Lgtm

PR #235, opened 2024-12-09
Update with new generator version

MatKuhr:
## Context

Applies [these changes](https://github.com/SAP/cloud-sdk-java/pull/657)

### DIFF
--- a/orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
+++ b/orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
@@ -11,13 +11,10 @@ import com.sap.ai.sdk.core.AiCoreService;
 import com.sap.ai.sdk.core.DeploymentResolutionException;
 import com.sap.ai.sdk.orchestration.model.CompletionPostRequest;
 import com.sap.ai.sdk.orchestration.model.CompletionPostResponse;
-import com.sap.ai.sdk.orchestration.model.FilterConfig;
 import com.sap.ai.sdk.orchestration.model.LLMModuleResult;
-import com.sap.ai.sdk.orchestration.model.MaskingProviderConfig;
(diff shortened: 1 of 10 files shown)


orchestration/src/main/java/com/sap/ai/sdk/orchestration/model/DataRepositoryType.java
```diff
@@ -20,8 +20,6 @@
 public enum DataRepositoryType {
   VECTOR("vector"),
 
-  HELP_SAP_COM("help.sap.com"),
```
MatKuhr, line 23: Side effect from not having run the generator on the latest spec yet

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
```diff
@@ -47,9 +44,6 @@ public class OrchestrationClient {
     JACKSON.addMixIn(
         ModuleResultsOutputUnmaskingInner.class,
         JacksonMixins.ModuleResultsOutputUnmaskingInnerMixIn.class);
```
CharlesDuboisSAP, line 46: Why are the mix ins `LLMModuleResult` and `ModuleResultsOutputUnmaskingInner` still there?
It means we are forcing interfaces to deserialize as `synchronous` classes, it can never be `streaming` classes

CharlesDuboisSAP (commented):
Other than my question LGTM

CharlesDuboisSAP (approved):

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
MatKuhr: Because https://github.tools.sap/AI/llm-orchestration/issues/778 and https://github.tools.sap/AI/llm-orchestration/issues/779 are not done yet, and deduction doesn't work for streaming

PR #243, opened 2024-12-18
chore: [Orchestration] Improve Streaming Sample Code

MatKuhr:
## Context

Gemini consistently refuses to generate enough numbers, and with the default chunk size of 100 barely any streaming occurs. The new prompt is more reliable


CharlesDuboisSAP (dismissed):

Jonas-Isr (approved):
Besides @newtork 's point this looks good to me!

Jonas-Isr (approved):

PR #246, opened 2024-12-19
chore: [DevOps] Update to Cloud SDK 5.15.0

MatKuhr:
## Context

Also re-enables the generators 🥳

### DIFF
--- a/core/src/main/java/com/sap/ai/sdk/core/model/InlineObject.java
+++ b/core/src/main/java/com/sap/ai/sdk/core/model/InlineObject.java
@@ -24,28 +24,28 @@ import java.util.Set;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
-/** ArtifactQuery400Response */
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/sap/ai/sdk/core/model/InlineObject.java
```diff
@@ -24,9 +24,9 @@
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
-/** FileDownload400Response */
+/** InlineObject */
 // CHECKSTYLE:OFF
-public class FileDownload400Response
```
MatKuhr, line 29: Not sure why this changed, but not worth investigating as these objects are unused anyway

newtork: Weird :/ 

> these objects are unused anyway

True.

newtork (approved):

PR #270, opened 2025-01-10
fix: Add actual messages to AI Core endpoints in sample app

Jonas-Isr:
## Context

AI/ai-sdk-java-backlog#147.

The sample app should give "useful" responses when clicking on the endpoints, regardless whether "JSON" or "Message" is selected in the footer. Before, when clicking the endpoints under AI Core while "Message" is selected, an error occurs. With this PR, some "useful" message are returned instead.

### Feature scope:
 
- [x] Add actual messages to the AI Core endpoints

## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~

### DIFF
--- a/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/ConfigurationController.java
+++ b/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/ConfigurationController.java
@@ -3,6 +3,7 @@ package com.sap.ai.sdk.app.controllers;
 import com.sap.ai.sdk.core.client.ConfigurationApi;
 import com.sap.ai.sdk.core.model.AiConfiguration;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestHeader;
@@ -23,7 +24,7 @@ class ConfigurationController {
    */
   @GetMapping("/configurations")
   ResponseEntity<Object> getConfigurations(
-      @RequestHeader(value = "accept", required = false) final String accept) {
+      @Nullable @RequestHeader(value = "accept", required = false) final String accept) {
     final var configList = CLIENT.query("default");
     if ("application/json".equals(accept)) {
       return ResponseEntity.ok().body(configList);
--- a/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/ConfigurationController.java
+++ b/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/ConfigurationController.java
@@ -1,10 +1,6 @@
 package com.sap.ai.sdk.app.controllers;
 
-import com.fasterxml.jackson.annotation.JsonAutoDetect;
-import com.fasterxml.jackson.annotation.PropertyAccessor;
 import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import com.sap.ai.sdk.core.client.ConfigurationApi;
 import com.sap.ai.sdk.core.model.AiConfiguration;
 import org.springframework.http.MediaType;
@@ -20,10 +16,6 @@ import java.util.stream.Collectors;
 class ConfigurationController {
 
   private static final ConfigurationApi CLIENT = new ConfigurationApi();
-  private static final ObjectMapper MAPPER =
-      new ObjectMapper()
-          .setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY)
-          .registerModule(new JavaTimeModule());
 
   /**
    * Get the list of configurations.
@@ -32,14 +24,14 @@ class ConfigurationController {
    * @return a response entity with a string representation of the list of configurations
    */
   @GetMapping("/configurations")
-  ResponseEntity<String> getConfigurations(
+  ResponseEntity<Object> getConfigurations(
       @RequestHeader(value = "accept", required = false) final String accept)
       throws JsonProcessingException {
     final var configList = CLIENT.query("default");
     if ("application/json".equals(accept)) {
       return ResponseEntity.ok()
           .contentType(MediaType.APPLICATION_JSON)
-          .body(MAPPER.writeValueAsString(configList));
+          .body(configList);
     }
     final var items =
         configList.getResources().stream()
--- a/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/ConfigurationController.java
+++ b/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/ConfigurationController.java
@@ -1,14 +1,13 @@
 package com.sap.ai.sdk.app.controllers;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.sap.ai.sdk.core.client.ConfigurationApi;
 import com.sap.ai.sdk.core.model.AiConfiguration;
+import java.util.stream.Collectors;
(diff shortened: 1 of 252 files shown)


sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/ConfigurationController.java
```diff
@@ -1,8 +1,17 @@
 package com.sap.ai.sdk.app.controllers;
 
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.PropertyAccessor;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+// import com.fasterxml.jackson.datatype.*;
```
newtork, line 7:
```suggestion
```

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/ConfigurationController.java
```diff
@@ -11,14 +20,42 @@
 class ConfigurationController {
 
   private static final ConfigurationApi CLIENT = new ConfigurationApi();
+  private final ObjectMapper mapper =
```
newtork, line 23: (Minor)
I think this can be static (allcaps)

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/ConfigurationController.java
```diff
@@ -11,14 +20,42 @@
...
+   * @param configList The configuration list.
+   * @return the message
+   */
+  private String buildMessage(final AiConfigurationList configList) {
+    final var message = new StringBuilder("The following configurations are available: ");
+    for (final var resource : configList.getResources()) {
+      message.append(resource.getName()).append(", ");
+    }
+    message.setCharAt(message.length() - 2, '.');
+    return message.toString();
```
newtork, line 59: (Minor)

Code can be reduced.

```suggestion
    final var items =
        configList.getResources().stream().map(AiConfiguration::getName).collect(joining(", "));
    return ResponseEntity.ok("The following configurations are available: %s.".formatted(items));
```

with
```
import static java.util.stream.Collectors.joining;
```

---

Google Commons library also has some nice convenience API. But I wouldn't advertise this in a sample app:
```
final var items =
  Joiner.on(", ").join(Iterables.transform(configList.getResources(), AiConfiguration::getName));
```

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/DeploymentController.java
```diff
@@ -52,62 +62,137 @@ AiDeploymentDeletionResponse createAndDeleteDeploymentByConfigId(
...
+    final var stoppedDeployments =
+        myDeployments.stream()
+            .map(
+                deployment ->
+                    CLIENT.modify(
+                        RESOURCE_GROUP,
+                        deployment.getId(),
+                        AiDeploymentModificationRequest.create()
+                            .targetStatus(AiDeploymentTargetStatus.STOPPED)))
+            .toList();
```
newtork, line 113: (Minor)
Readability can be improved, I think.

```suggestion
    final var modificationRequest = AiDeploymentModificationRequest.create().targetStatus(STOPPED);
    final var stoppedDeployments =
        myDeployments.stream()
            .map(AiDeployment::getId)
            .map(id -> CLIENT.modify(RESOURCE_GROUP, id, modificationRequest))
            .toList();
```

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/DeploymentController.java
```diff
@@ -10,38 +15,43 @@
...
+import org.springframework.web.bind.annotation.RequestHeader;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 
 /** Endpoints for AI Core AiDeployment operations */
 @Slf4j
 @RestController
+@SuppressWarnings("unused")
 @RequestMapping("/deployments")
 class DeploymentController {
```
newtork, line 38: (Comment)

Too bad we have to add this annotation.
Normally IntelliJ default settings are clever enough to identify program / service entry points.
I don't know any other solution.

newtork (commented):
Only minor comments.

Also I've noticed the warnings about input-param used in output. I actually would have suggested that too, what you had before. But if it was flagged as warning, I'm fine with current static String responses :)

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/DeploymentController.java
```diff
@@ -52,62 +55,109 @@ AiDeploymentDeletionResponse createAndDeleteDeploymentByConfigId(
...
+   * @param accept The accept header.
+   * @return a response entity with a string representation of the deployment creation response
+   */
+  @GetMapping("/by-config/{id}/createDelete")
+  ResponseEntity<Object> createAndDeleteDeploymentByConfigId(
+      @Nonnull @PathVariable("id") final String configId,
+      @RequestHeader(value = "accept", required = false) final String accept) {
+    final var response = createAndDeleteDeploymentByConfigId(configId);
+    if ("application/json".equals(accept)) {
+      return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(response);
```
rpanackal, line 70: Comment:

I found that we don't need to explicitly set the media type header in response. Spring can automatically figure it from the request header and calls the corresponding "converter" internally.

`ResponseEntity.ok().body(response);` will produce a valid json as long as request accept header is `application/json`. It also sets the response `accept` header to `application/json`. Similarly, for `text/html`.

In case the type of response is not meant to be deserialised, we can pass it as a string with a custom mapper like `ResponseEntity.ok().body(mapper.writeValueAsString(response));`.

I could replicate the behaviour with our sample app but couldn't find any docs to support linking the cause of this behaviour to `accept` header.

[Related doc](https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-controller/ann-methods/responseentity.html)

Jonas-Isr: Removed the unnecessary `.contentType(MediaType.APPLICATION_JSON)` call.

I do not understand your last point about a link to the `accept` header.

rpanackal: I am no spring expert.

I tried out what happens without the media type being explicitly set on `ResponseEntity` and seems to work correctly. 

Still, take it with a pinch of salt, as I failed to find any clear doc stating that this works because spring uses the accept header in the request to infer the response representation.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/DeploymentController.java
```diff
@@ -52,62 +54,109 @@ AiDeploymentDeletionResponse createAndDeleteDeploymentByConfigId(
...
+   * @return a response entity with a string representation of the deployment modification response
    */
   @GetMapping("/by-config/{id}/stop")
   @Nonnull
-  @SuppressWarnings("unused") // debug method that doesn't need to be tested
-  List<AiDeploymentModificationResponse> stopByConfigId(
-      @Nonnull @PathVariable("id") final String configId) {
+  ResponseEntity<Object> stopByConfigId(
+      @Nonnull @PathVariable("id") final String configId,
+      @RequestHeader(value = "accept", required = false) final String accept) {
```
newtork, line 87: (MInor)

```suggestion
      @Nullable @RequestHeader(value = "accept", required = false) final String accept) {
```

Applies to other methods as well.

newtork (approved):

PR #273, opened 2025-01-13
chore: [DevOps] PR Title Linter

MatKuhr:
Examples which are allowed:

* `chore: [DevOps] PR Title Linter`
* `chore: [DevOps] pr title linter`
* `chore: PR Title Linter`

Examples which are not allowed:

* `PR Title Linter`
* `[DevOps] PR Title Linter`
* `chore: [Foo] PR Title Linter`

[Tested on my fork](https://github.com/MatKuhr/ai-sdk-java/actions/runs/12743754869/job/35514342688?pr=2)

### DIFF
--- a/.github/workflows/pr-lint.yaml
+++ b/.github/workflows/pr-lint.yaml
@@ -23,7 +23,7 @@ jobs:
             OpenAI
             Orchestration
             DevOps
-          headerPattern: '^(\w+): (?:\[(\w+)\] )?(.+)$'
+          headerPattern: '^(\w.+): (?:\[(\w.+)\] )?(.+)$'
           headerPatternCorrespondence: type, scope, subject
           # for available types, check:
           # https://github.com/commitizen/conventional-commit-types/blob/master/index.json
--- a/.github/workflows/pr-lint.yaml
+++ b/.github/workflows/pr-lint.yaml
@@ -1,0 +1,33 @@
+name: lint pr
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+permissions:
+  pull-requests: read
+
+jobs:
+  lint-pr:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          scopes: |
+            Core
+            OpenAI
+            Orchestration
+            DevOps
+          headerPattern: '^(\w+): (\[\w+\] )?(.+)$'
+          headerPatternCorrespondence: type, scope, subject
+          # for available types, check:
+          # https://github.com/commitizen/conventional-commit-types/blob/master/index.json
+          subjectPatternError: |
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-lint.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          scopes: |
+            Core
+            OpenAI
+            Orchestration
+            DevOps
+          headerPattern: '^(\w+): (?:\[(\w+)\] )?(.+)$'
```
newtork, line 26: (Minor)

I would like a `-`, e.g. when refering to `[spring-app]`

```suggestion
          headerPattern: '^(\w+): (?:\[([\w-]+)\] )?(.+)$'
```

.github/workflows/pr-lint.yaml
```diff
@@ -0,0 +1,33 @@
+name: lint pr
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
```
newtork, line 8: (Comment/Opinion)

I didn't expect this syntax to be mandatory.
* It may discourage external contributors.
* It may break scripts and automated PRs.

It's not possible to make this more like a warning/notification right?

.github/workflows/pr-lint.yaml
MatKuhr: This regex only defines the capture groups. The allowed topics are (currently) hard coded in the list above. We can use regex there as well, but I felt it isn't needed

.github/workflows/pr-lint.yaml
MatKuhr: > more like a warning

No, at least not per the docs. I checked that it works with Dependabot, but Release PRs would currently fail. Will update the script. For external contributors it should be fine, because it is easy to fix and only occurs after a PR is already raised, i.e. when most of the work is already done.

.github/workflows/pr-lint.yaml
newtork: Interesting... Do I understand correctly [according to the readme](https://github.com/amannn/action-semantic-pull-request), the users of this github action are required to define an allowlisted set of `scopes`?

MatKuhr: I think defining scopes is optional, didn't test it though. But since scopes themselves are already configured to be optional, I figured we can use a few defined values to prevent `Core` `AiCore` `AICore` `Ai-Core` ...

newtork: Good idea. Though, without having checked that, I doubt this will work as expected(?)
`[Ai-Core]` will just be part of third capture group, leaving second group empty. Maybe.

MatKuhr:
```suggestion
          headerPattern: '^(\w.+): (?:\[(\w.+)\] )?(.+)$'
```

True, fixed it now ✅

CharlesDuboisSAP (approved):

PR #282, opened 2025-01-16
chore: [DevOps] Automation to Update Spec Files

MatKuhr:
## Context

Automation to update spec files.

Features:

- One job to rule them all
- Supports spec version based on tag or branch
- Can be triggered from outside this repo as well
- Creating a PR is optional, which is useful for testing

### DIFF
--- a/.github/workflows/spec-update.yaml
+++ b/.github/workflows/spec-update.yaml
@@ -1,0 +1,140 @@
+name: "Spec File Update Workflow"
+
+on:
+  push:
+    branches:
+      - chore/spec-update-job
+
+  workflow_dispatch:
+    inputs:
+      file:
+        description: "Which spec file to update"
+        type: choice
+        required: false
+        options:
+          - core
+          - grounding
+          - orchestration
+        default: orchestration
+      file-ref:
+        description: "Branch or tag to checkout the spec file from"
+        required: false
+        default: main
+        type: string
+      create-pr:
+        description: "Create a pull request after updating the spec file"
+        required: false
+        default: true
+        type: boolean
+
+env:
+  MVN_MULTI_THREADED_ARGS: --batch-mode --no-transfer-progress --fail-at-end --show-version --threads 1C
+  JAVA_VERSION: 17
+
+jobs:
+  generate:
+    name: "Download, Generate, Compile and Push"
+    runs-on: [ubuntu-latest]
+    permissions:
+      pull-requests: write
+      contents: write
+    outputs:
+      spec_diff: ${{ steps.spec_diff.outputs.spec_diff }}
+      branch: ${{ steps.push.outputs.branch }}
+    env:
+      API_BASE_URL: "https://github.tools.sap/api/v3/repos"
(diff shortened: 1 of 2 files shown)


.github/workflows/spec-update.yaml
```diff
@@ -0,0 +1,180 @@
...
+          token: ${{ secrets.BOT_SDK_JS_FOR_DOCS_REPO_PR }}
+
+      - name: "Download specification file"
+        id: download
+        env:
+          GH_ENTERPRISE_TOKEN: ${{ secrets.GH_TOOLS_TOKEN }}
+        run: |
+          case $CHOICE in
+            core)
+              API_URL="$API_BASE_URL/cloudsdk/cloud-sdk-java-tests/contents/aicore.yaml?ref=$REF"
```
CharlesDuboisSAP, line 60: This path doesn't exist.
Also could you run the script for AI Core once?

.github/workflows/spec-update.yaml
```diff
@@ -0,0 +1,180 @@
...
+              FILE_PATH='core/src/main/resources/spec/aicore.yaml'
+              ;;
+            grounding)
+              # TODO
+              exit 1
+              ;;
+            orchestration)
+              API_URL="$API_BASE_URL/AI/llm-orchestration/contents/src/spec/api.yaml?ref=$REF"
```
newtork, line 68: (Major/Question)

I was told we should not use internal specification file, and there would be a "manually refined" spec file version for public purpose.

Has this assumption changed?

.github/workflows/spec-update.yaml
MatKuhr: https://github.tools.sap/cloudsdk/cloud-sdk-java-tests/blob/main/aicore.yaml

Will do after merge, because without the manual trigger I always have to hard-code the params

.github/workflows/spec-update.yaml
```diff
@@ -0,0 +1,180 @@
...
+          if mvn test-compile ${{ env.MVN_MULTI_THREADED_ARGS }} ; then
+            echo "compilation_result=success" >> "$GITHUB_OUTPUT"
+            if mvn test ${{ env.MVN_MULTI_THREADED_ARGS }} ; then
+                echo "test_result=success" >> "$GITHUB_OUTPUT"
+            else
+                echo "test_result=failure" >> "$GITHUB_OUTPUT"
+            fi
+          else
+            echo "compilation_result=failure" >> "$GITHUB_OUTPUT"
+          fi
```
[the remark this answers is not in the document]
MatKuhr: It's useful for cases where we only want to test a spec update without creating a PR. This is intended for testing a spec file from a branch in development. I could also trigger the CI workflow, but that might be overkill.

.github/workflows/spec-update.yaml
MatKuhr: Yes, currently the idea is that this file can be used as-is for publishing. But not decided yet, and this can easily be changed to point to a different file.

newtork: Hopefully we remember to change this line once a (different) decision was made.

newtork (approved):

PR #294, opened 2025-01-23
feat: Orchestration image support

Jonas-Isr:
## Context

AI/ai-sdk-java-backlog#152.

We want to support using images and prompts with multiple text messages as input for Orchestration. 

### Feature scope:
 
- [x] Added possibility of using image and multiple text messages in prompts
- [x] Added Convenience layer

## Definition of Done

- [x] Documentation updated
- [x] Release notes updated


orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
```diff
@@ -46,8 +55,96 @@ static SystemMessage system(@Nonnull final String msg) {
...
+                                  + multiMessageContent.getClass());
+                        }
+                      })
+                  .toList());
+    } else {
+      throw new IllegalArgumentException("Unknown content type: " + this.getContent().getClass());
+    }
+  }
+
+  static Message addTextMessages(
```
Jonas-Isr, line 92: Question: Is there a better way to distribute the logic of `addTextMessages()` and `addImage()` between the Message interface and its implementing classes?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
```diff
@@ -1,17 +1,58 @@
...
+    return MessageContent.toString(content);
+  }
+
+  @Override
+  @Nonnull
+  public MessageContent getContent() {
+    return content != null ? content : new MessageContentSingle("");
+  }
+
+  public UserMessage(String singleMessage) {
```
Jonas-Isr, line 34: Comment: I added these constructors for backwards compatibility and convenience. To enable dynamic message building, I added the `addTextMessage()` and `addImage()` methods. Instead, we could use a builder pattern, but I did not find a way to make that work without breaking changes to the constructors.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/MessageContent.java
```diff
@@ -0,0 +1,29 @@
+package com.sap.ai.sdk.orchestration;
+
+public sealed interface MessageContent permits MessageContentSingle, MessageContentMulti {
+
+  static String toString(MessageContent content) {
```
Jonas-Isr, line 5: Question: This is to turn the message content into a string. Is there a more elegant way/place to do this?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/MessageContentMulti.java
```diff
@@ -0,0 +1,36 @@
...
+import com.sap.ai.sdk.orchestration.model.ImageContent;
+import com.sap.ai.sdk.orchestration.model.MultiChatMessageContent;
+import com.sap.ai.sdk.orchestration.model.TextContent;
+import java.util.List;
+
+public record MessageContentMulti(List<MultiMessageContent> multiContentList)
+    implements MessageContent {
+  public MessageContentMulti {}
+
+  public MessageContentMulti(MultiChatMessageContent... multiChatContents) {
```
Jonas-Isr, line 14: Question: I use `MultiChatMessageContent...` as input here since using `List<MultiChatMessageContent>` results in a compilation error ("constructors have same erasure"). Is there a more elegant way to solve this?

Jonas-Isr:
Comment: I added 4 new classes and 2 new interfaces to basically mirror the generated classes that hold the new types of message-content:
 - `MessageContent` is an interface which appears in the `content` field of a `Message` and is used as an umbrella for the two new classes `MessageContentSingle` and `MessageContentMulti`. The first one is used in the `content` field of messages that only hold a single string as context (the *standard* messages, so to say). The second one is for messages that either have an image or multiple strings as content.
 - A `MessageContentMulti` object holds a list of `MultiMessageContent`. This mirrors the list of texts and images the `Message` object is holding.
 - `MultiMessageContent` is an interface which is used as an umbrella for the two classes `MultiMessageTextContent` and `MultiMessageImageContent`. These then either hold a string (for the text) or a string and an enum detailLevel (for the image URL and its detail level).

I would be super happy to find better names for these new classes and interfaces :D

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
```diff
@@ -1,17 +1,58 @@
...
+  }
+
+  public UserMessage(List<MultiChatMessageContent> multiChatMessageContentList) {
+    content =
+        new MessageContentMulti(
+            multiChatMessageContentList.toArray(MultiChatMessageContent[]::new));
+  }
+
+  @Nonnull
+  public UserMessage addTextMessages(@Nonnull String... messages) {
```
newtork, line 49: (Minor)

Maybe for API consistency we could make it singular (similar to image) and drop the "message" suffix?

```suggestion
  public UserMessage addText(@Nonnull String message) {
```

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
```diff
@@ -18,6 +23,17 @@ static UserMessage user(@Nonnull final String msg) {
...
   }
 
+  /**
+   * A convenience method to create a user message.
+   *
+   * @param msg the message content.
+   * @return the user message.
+   */
+  @Nonnull
+  static UserMessage user(@Nonnull final MessageContent msg) {
```
Jonas-Isr, line 33: I added new convenience methods to create user and system messages using a `MessageContent` object. Should we have something similar for assistant messages for consistency (even though we only allow a single message there?).

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
```diff
@@ -1,58 +1,48 @@
...
-  }
+  @Nonnull MessageContent content;
 
   @Override
   @Nonnull
-  public MessageContent getContent() {
-    return content != null ? content : new MessageContentSingle("");
+  public MessageContent content() {
+    return content;
   }
```
newtork, line 24: (MInor)

You since we have fluent lombok accessors on this class, you can omit this method declaration. I expect it to be generated automatically.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
newtork: Maybe you showed this in another place, but how is the developer expected to interact with this static factory method, compared to when not-having this method? What is the benefit? Just wondering whether it's worth to having additional public API reference to `MessageContent` here.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
```diff
@@ -49,16 +56,29 @@ static SystemMessage system(@Nonnull final String msg) {
...
    * Converts the message to a serializable ChatMessage object.
    *
    * @return the corresponding {@code ChatMessage} object.
    */
   @Nonnull
   default ChatMessagesInner createChatMessage() {
-    if (this.getContent() instanceof MessageContentSingle) {
-      return ChatMessage.create().role(role()).content(content());
-    } else if (this.getContent() instanceof MessageContentMulti mCMulti) {
+    if (this.content() instanceof MessageContentSingle) {
```
newtork, line 77: (Minor/Opinion)

I think it's a mistake to have this method on public API.
Maybe we could address this in another PR.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/MessageContentMulti.java
newtork: * Let's check why / whether this class needs to be public.
* I would not offer the `(newContent, oldContent)` constructor in the first place. We can expect the developer to merge it on consumption side.
* You could change the type from `List` to `Iterable`, whcih allows you to use guava convenience method `listMerged = Iterables.concat(list1, list2);`

orchestration/src/main/java/com/sap/ai/sdk/orchestration/SystemMessage.java
```diff
@@ -15,42 +14,37 @@ public class SystemMessage implements Message {
...
+      ((MessageContentMulti) messageContent)
+          .multiContentList().stream()
+              .filter(mCMC -> !(mCMC instanceof MultiMessageTextContent))
+              .findAny()
+              .ifPresent(
+                  mCMC -> {
+                    throw new IllegalArgumentException(
+                        "Only TextContent is supported for SystemMessage");
+                  });
+    }
```
newtork, line 40: (Minor)
I would leave it out. Prematurely throwing exceptions at runtime is an unnecessary risk. There is no benefit for developers at design time right? The user would still do the regular casts.
Instead I would just add a javadoc or comment, that only instances of `MultiMessageTextContent` are expected.

Jonas-Isr:
After thinking about this some time (and switching back and forth in my opinion) I am not sure whether it makes more sense to delete the class `MessageContentSingle` and make `MessageContent` a class that simply holds a list of `MultiMessageContent`s. That would get rid of a bit of complexity and the need to cast things around.

What do you think? (We can talk about this in the meeting later.)

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
newtork: > I added this mainly because I wanted to give the user the opportunity to create a user message that only consists of an image

I understand. But this is why I was asking for expectations in consuming the API. 

* The developer see's `MessageContent` and realized they can't instantiate via constructor nor factory method. _(Problem)_

* The developer would need to browse the code to find `MessageContentSingle` only offers `String`, and the second implementation `MessageContentMulti` (while being counter intuitive for single image) has another class reference `MultiChatMessageContent`. 

* The developer again needs to figure out how to instantiate an object. Only while looking up classes will find `MultiMessageImageContent`.

newtork (commented):
I want to explain my assumptions for convenience API behavior / contract.

* For request:
  * the user will only use public constructors and factory methods for easy instantiation.
  * low API hierarchy requirement; we'll try to avoid users to define helper classes.
  * strict value checking on design time; only allow what's aligned in spec.

* For responses:
  * developer uses public accessors.
  * while casting may be necessary, let's try to minimalize the effort.
  * hierarchy can be high as long as it's reasonable.
  * loose value checking; what ever is sent back from server can be read with our API.

Jonas-Isr:
The convenience API for constructing messages is as follows right now:

```
//    Creation of messages (same for SystemMessage)
    var message = Message.user("Hello!");                              // also for AssistantMessage
    var message = Message.user("Hello!", "How ar you?");
    var message = Message.user(MessageContent.text("Hello!"));
    var message = Message.user(MessageContent.text("Hello!", "How ar you?"));
    var message = new UserMessage("Hello!");                            //also for AssistantMessage
    var message = new UserMessage("Hello!", "How ar you?");
    var message = new UserMessage(MessageContent.text("Hello!"));
    var message = new UserMessage(MessageContent.text("Hello!", "How ar you?"));

//    Adding text (same for SystemMessage)
    message.addText("How are you?");
    message.addText("How are you?", "What's up?'");
    message.add(MessageContent.text("How are you?"));
    message.add(MessageContent.text("How are you?", "What's up?'"));

//    Adding image (only for UserMessage)
    message.addImage("url");
    message.addImage("url", ImageItem.DetailLevel.low);
    message.add(MessageContent.image("url"));
    message.add(MessageContent.image("url", ImageItem.DetailLevel.low));
```

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
```diff
@@ -66,5 +121,5 @@ default ChatMessage createChatMessage() {
    */
   @Nonnull
   @Beta
-  String content();
+  MessageContent content();
```
Jonas-Isr, line 124: This is a **breaking change**. (Note that the corresponding methods in `UserMessage` etc. were not marked as beta.)

We discussed in the meetings that we do not want to offer a convenience method that returns a string representation of messages anymore. Users can/will use `response.getContent()` for nearly all use cases and it is not clear how a sensible string representation for messages with many `TextItem`s or `IamgeItem`s should look like.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/SystemMessage.java
```diff
@@ -13,5 +16,58 @@ public class SystemMessage implements Message {
...
+      @Nonnull final String message, @Nullable final String... additionalMessages) {
+    content = MessageContent.text(message, additionalMessages);
+  }
+
+  /**
+   * Creates a new system message with the given message content.
+   *
+   * @param messageContent the message content.
+   */
+  public SystemMessage(@Nonnull final MessageContent messageContent) {
```
Jonas-Isr, line 37: We do not check here whether the `MessageContent` object contains `ImageItem`s, even though no LLM currently allows system messages with images. 
We decided against it in a meeting because this might change at any moment and additionally, even if we would throw an exception here, the user would not get any more convenience out of that compared to the call being rejected by an LLM.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
```diff
@@ -48,7 +79,31 @@ static SystemMessage system(@Nonnull final String msg) {
    */
   @Nonnull
   default ChatMessage createChatMessage() {
-    return SingleChatMessage.create().role(role()).content(content());
+    final var itemList = this.content().contentItemList();
+    if (itemList.size() == 1 && itemList.get(0) instanceof TextItem textItem) {
+      return SingleChatMessage.create().role(role()).content(textItem.text());
```
Jonas-Isr, line 84: Here we cannot discern between `Message.user("text")` and `Message.user(MessageContent.text("text"))`, so both get turned into `SingleChatMessage`. Is that a problem?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/MessageContent.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.stream.Stream;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Represents the content of a chat message.
+ *
+ * @param contentItemList a list of the content items
+ */
+public record MessageContent(@Nonnull List<ContentItem> contentItemList) {
```
Jonas-Isr, line 17: Currently, it is possible to create a `MessageContent` object with an empty list, since for example GPT4o allows to call it with messages that have `"content" : []`. Others, like Gemini, don't allow such messages.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
```diff
@@ -2,43 +2,74 @@
...
 
+  /**
+   * A convenience method to create a user message from one or more strings.
+   *
+   * @param message the message content.
+   * @param additionalMessages the additional messages.
+   * @return the user message.
+   */
+  @Nonnull
+  static UserMessage user(
```
MatKuhr, line 24: Why does user permit varargs but the others don't?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
MatKuhr: Users can use the constructors instead should this ever matter, right?

docs/guides/ORCHESTRATION_CHAT_COMPLETION.md
```diff
@@ -259,6 +259,39 @@ try (Stream<String> stream = client.streamChatCompletion(prompt, config)) {
...
+
+It's possible to add images and multiple text inputs to a message.
+
+### Add images to a message
+
+There are multiple ways to add an image to a user message as detailed below. Note, that currently only user messages are supported for image attachments.
+
+```java
+var message1 = Message.user("What is in this image?").addImage("https://url.to/image.jpg");
+var message2 = Message.user("And what is in this image?").add(MessageContent.image("https://url.to/image.jpg"));
```
MatKuhr, line 273: (minor) Users may wonder why we have two examples here, and if they are different in any way. Personally, I'd KISS and only give one example.

What is not obvious here is if a message can contain only an image, without any text

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
Jonas-Isr: It is also allowed for `system()` (see Line 58 below). It is not allowed for assistant messages since they can/should only hold a single text.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
Jonas-Isr: The constructor is protected, but the `create()` methods etc. are public, so yes.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
MatKuhr: While currently the spec limits response messages to single, users are technically allowed (and it works with GPT models) to pass multi assistant messages. But I'd say for now we can just not allow this on the convenience API.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/SystemMessage.java
```diff
@@ -13,5 +16,58 @@ public class SystemMessage implements Message {
...
+  }
+
+  /**
+   * Add text to the message.
+   *
+   * @param messages the text to add.
+   * @return the new message.
+   */
+  @Nonnull
+  public SystemMessage addText(@Nonnull final String... messages) {
```
MatKuhr, line 48: while immutability can be nice, the verb `add` kinda suggests a mutating behavior.

e.g. if I run:

```java
var msg = Message.system("Foo");
msg.addText(" and Bar");
System.out.println(msg);
```

this printing just `Foo` feels unexpected. Maybe we find a better naming, e.g. `andText` which sounds more like a builder..

docs/release-notes/release_notes.md
```diff
@@ -8,11 +8,12 @@
 
 ### 🔧 Compatibility Notes
 
--
+- `Message.content()` returns a `ContentItem` now instead of a `String`. Use `((TextItem) Message.content().contentItemList().get(0)).text()` if the corresponding `ContentItem` is a `TextItem` and the string representation is needed.
```
MatKuhr, line 11:
```suggestion
- `Message.content()` returns a `ContentItem` now instead of a `String`. Use `((TextItem) Message.textContent()` if the corresponding `ContentItem` is a `TextItem` and the string representation is needed.
```
maybe we can hide this behind a (potentially beta) convenience method?

Jonas-Isr: Yes, sounds good :)
Should I then simply divide the different strings with a semicolon or something? And for `ImageContent` I print the url?

MatKuhr: I think we can just throw, which is what would happen in the type cast anyway

Jonas-Isr: Discussed that we do not want to enable this.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ImageItem.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  /**
+   * Creates a new image item with the given image URL.
+   *
+   * @param imageUrl the URL of the image
+   */
+  public ImageItem(@Nonnull final String imageUrl) {
+    this(imageUrl, DetailLevel.auto);
+  }
```
MatKuhr, line 34: These two seem redundant, why would we need `new ImageItem(url, null)` if we have `new ImageItem(url)`?  (also applies to `MessageContent` & `UserMessage`)

orchestration/src/test/java/com/sap/ai/sdk/orchestration/OrchestrationUnitTest.java
```diff
@@ -691,132 +678,142 @@ void streamChatCompletionDeltas() throws IOException {
...
+            .willReturn(aResponse().withStatus(SC_OK).withBodyFile("multiMessageResponse.json")));
+
+    OrchestrationAiModel customGpt4o =
+        GPT_4O_MINI
+            .withParam(MAX_TOKENS, 50)
+            .withParam(TEMPERATURE, 0.1)
+            .withParam(FREQUENCY_PENALTY, 0)
+            .withParam(PRESENCE_PENALTY, 0)
+            .withParam(TOP_P, 1)
+            .withParam(N, 1);
```
MatKuhr, line 693: why set these for this test?

orchestration/src/test/java/com/sap/ai/sdk/orchestration/OrchestrationUnitTest.java
```diff
@@ -691,132 +678,142 @@ void streamChatCompletionDeltas() throws IOException {
...
-                .registerModule(module)
-                .readValue(
-                    new String(
-                        fileLoader.apply("__files/multiChatMessageResponse.json").readAllBytes()),
-                    CompletionPostResponse.class));
-
-    assertThatThrownBy(orchestrationChatResponse::getAllMessages)
-        .isInstanceOf(UnsupportedOperationException.class)
-        .hasMessage("Messages of MultiChatMessage type not supported by convenience API");
+  void testMessageConstruction() {
```
MatKuhr, line 792: good to cover this with a test 👍🏻 

If you like, you could also move this to a separate test file, since it doesn't need the client and this test is already long 😄

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OrchestrationService.java
```diff
@@ -53,6 +62,45 @@ public OrchestrationChatResponse completion(@Nonnull final String famousPhrase)
...
+  @Nonnull
+  public OrchestrationChatResponse imageInput(@Nonnull final String pathToImage) {
+    final OrchestrationAiModel customGpt4o =
+        GPT_4O_MINI
+            .withParam(MAX_TOKENS, 50)
+            .withParam(TEMPERATURE, 0.1)
+            .withParam(FREQUENCY_PENALTY, 0)
+            .withParam(PRESENCE_PENALTY, 0)
+            .withParam(TOP_P, 1)
+            .withParam(N, 1);
```
MatKuhr, line 79: unless we really need this for this test case, I'd rather keep the focus of the sample code on the image input

MatKuhr (dismissed):
otherwise lgtm 👍🏼

docs/release-notes/release_notes.md
```diff
@@ -8,11 +8,12 @@
...
 
 ### ✨ New Functionality
 
 - [Add Orchestration `LlamaGuardFilter`](../guides/ORCHESTRATION_CHAT_COMPLETION.md#chat-completion-filter).
+- [Convenient methods to create messages containing images and multiple text inputs](../guides/ORCHESTRATION_CHAT_COMPLETION.md#add-images-and-multiple-text-inputs-to-a-message)
```
CharlesDuboisSAP, line 16:
```suggestion
- Orchestration:
  - [Add `LlamaGuardFilter`](../guides/ORCHESTRATION_CHAT_COMPLETION.md#chat-completion-filter).
  - [Convenient methods to create messages containing images and multiple text inputs](../guides/ORCHESTRATION_CHAT_COMPLETION.md#add-images-and-multiple-text-inputs-to-a-message)
```

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
```diff
@@ -13,5 +18,90 @@ public class UserMessage implements Message {
...
+  MessageContent content;
+
+  /**
+   * Creates a new user message from one or more strings.
+   *
+   * @param message the first message.
+   * @param additionalMessages the additional messages.
+   */
+  public UserMessage(@Nonnull final String message, @Nullable final String... additionalMessages) {
+    content = MessageContent.text(message, additionalMessages);
```
newtork, line 32: (Minor)
Please use constructor chaining: `this(MessageContent.text(message, additionalMessages));`

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
```diff
@@ -13,5 +18,90 @@ public class UserMessage implements Message {
...
+  }
+
+  /**
+   * Creates a new user message with the given message content.
+   *
+   * @param messageContent the message content.
+   */
+  public UserMessage(@Nonnull final MessageContent messageContent) {
```
newtork, line 40: (Minor)

Please omit default constructors that are generated by lombok.

Since, lombok notices a custom constructor above, you would need to mark it with `@Tolerate`, otherwise we'll not have the default constructor via `@Value`

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
```diff
@@ -13,5 +18,90 @@ public class UserMessage implements Message {
...
+   */
+  @Nonnull
+  public UserMessage andImage(
+      @Nonnull final String imageUrl, @Nonnull final ImageItem.DetailLevel detailLevel) {
+    return new UserMessage(
+        new MessageContent(
+            Stream.concat(
+                    content.contentItemList().stream(),
+                    Stream.of(new ImageItem(imageUrl, detailLevel)))
+                .toList()));
```
newtork, line 74: (Minor)

Please declare extra variable.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/MessageContent.java
```diff
@@ -0,0 +1,91 @@
...
+   */
+  @Nonnull
+  public static MessageContent text(
+      @Nonnull final String firstInput, @Nullable final String... moreInputs) {
+    return new MessageContent(
+        Stream.concat(
+                Stream.of(firstInput),
+                moreInputs == null ? Stream.empty() : Arrays.stream(moreInputs))
+            .map(text -> (ContentItem) new TextItem(text))
+            .toList());
```
newtork, line 64: (Minor)

Please simplify

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ContentItem.java
```diff
@@ -0,0 +1,4 @@
+package com.sap.ai.sdk.orchestration;
+
+/** Represents an item in a {@link MessageContent} object. */
+public sealed interface ContentItem permits TextItem, ImageItem {}
```
newtork, line 4: (Minor)

Please make sure to add `@since 1.3.0` on all new public API (new public classes and new public methods on old classes)

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
```diff
@@ -13,5 +18,90 @@ public class UserMessage implements Message {
...
+  }
+
+  /**
+   * Add content in the form of {@link MessageContent} objects to the message.
+   *
+   * @param messageContents the content to add.
+   * @return the new message.
+   */
+  @Nonnull
+  public UserMessage and(@Nonnull final MessageContent... messageContents) {
```
newtork, line 98: (Major/Question)

As a user, when would I use `and(MessageContent)`?
`ContentItem` - would understand but this above confuses me.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
```diff
@@ -2,43 +2,74 @@
...
+
+  /**
+   * A convenience method to create a system message from one or more strings.
+   *
+   * @param message the message content.
+   * @return the system message.
+   */
+  @Nonnull
+  static SystemMessage system(
+      @Nonnull final String message, @Nullable final String... additionalMessages) {
```
newtork, line 59: (Major/Question)

As a user, when would I prefer to use the varargs like `system(String, String, String)`?

I find the following use case unrealistic:
```java
var msg = Message.system(
    "You are a helpful chatbot that answers my questions...", // hundreds of words
    "You have a German accent while responding, try to make it satirical...", // another hundreds of words
    "Respond in ALL CAPS, make it intense"
) ;
```
 

More realistically the developer uses variables right?
```java
var msg = Message.system(MSG_BOT, MSG_GERMAN, MSG_ALLCAPTS);
```

In that case I don't see a benefit over proper API:
```java
var msg = Message.system(MSG_BOT).andText(MSG_GERMAN).andText(MSG_ALLCAPS);
```

I would avoid two public APIs offering similar result.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
```diff
@@ -48,7 +79,31 @@ static SystemMessage system(@Nonnull final String msg) {
...
+                                      .url(imageItem.imageUrl())
+                                      .detail(imageItem.detailLevel().toString()));
+                        } else {
+                          return TextContent.create()
+                              .type(TextContent.TypeEnum.TEXT)
+                              .text(((TextItem) contentItem).text());
+                        }
+                      })
+                  .toList());
+    }
```
newtork, line 106: (Minor)

Not a fan of the code above. The amount of indentations have a code smell.

<details><summary>No need really to use Stream API</summary>

```suggestion
    }
    final var items = new LinkedList<MultiChatMessageContent>();
    for (ContentItem contentItem : itemList) {
      if (contentItem instanceof TextItem textItem) {
        items.add(TextContent.create().type(TextContent.TypeEnum.TEXT).text(textItem.text()));
      } else if (contentItem instanceof ImageItem imageItem) {
        final var detail = imageItem.detailLevel().toString();
        final var img = ImageContentImageUrl.create().url(imageItem.imageUrl()).detail(detail);
        items.add(ImageContent.create().type(ImageContent.TypeEnum.IMAGE_URL).imageUrl(img));
      }
    }
    return MultiChatMessage.create().role(role()).content(items);
```

</details>

orchestration/src/main/java/com/sap/ai/sdk/orchestration/MessageContent.java
```diff
@@ -0,0 +1,91 @@
...
+          if (content instanceof TextContent text) {
+            return new TextItem(text.getText());
+          } else {
+            final var imageUrl = ((ImageContent) content).getImageUrl();
+            return new ImageItem(
+                imageUrl.getUrl(), ImageItem.DetailLevel.fromString(imageUrl.getDetail()));
+          }
+        };
+    return Arrays.stream(multiChatContents).map(convertMultiChatContent).toList();
+  }
```
newtork, line 90: (Minor)

Code can be reduced significantly.

<details><summary>Guava offers a <code>Lists.transform</code> helper that can work with lambdas and method references</summary>

```suggestion
    this(Lists.transform(Arrays.asList(multiChatContents), MessageContent::toContentItem));
  }

  @Nonnull
  private static ContentItem toContentItem(@Nonnull final MultiChatMessageContent content) {
    if (content instanceof TextContent text) {
      return new TextItem(text.getText());
    }
    final var imageUrl = ((ImageContent) content).getImageUrl();
    return new ImageItem(imageUrl.getUrl(), ImageItem.DetailLevel.fromString(imageUrl.getDetail()));
  }
```

</details>

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
```diff
@@ -13,5 +18,90 @@ public class UserMessage implements Message {
...
+  }
+
+  /**
+   * Add text to the message.
+   *
+   * @param messages the text to add.
+   * @return the new message.
+   */
+  @Nonnull
+  public UserMessage andText(@Nonnull final String... messages) {
```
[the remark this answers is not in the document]
Jonas-Isr: Yes, that is true. I started with `addText()` etc., but Matthias rightfully pointed out that this would imply mutability.

I have no strong personal opinion on this (and way less experience than you guys) so I am happy to change it. To align with everyone, should I start a discussion/vote in Slack?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationChatResponse.java
```diff
@@ -51,30 +53,45 @@ public TokenUsage getTokenUsage() {
...
+        messages.add(
+            switch (mCMessage.getRole()) {
+              case "user" ->
+                  Message.user(
+                      new MessageContent(
+                          mCMessage.getContent().toArray(MultiChatMessageContent[]::new)));
+              case "system" ->
+                  Message.system(
+                      new MessageContent(
+                          mCMessage.getContent().toArray(MultiChatMessageContent[]::new)));
```
newtork, line 83: (Minor)

This can be reduced a bit by avoiding duplicate code.

<details><summary>Example</summary>

```suggestion
        final var contents = mCMessage.getContent().toArray(MultiChatMessageContent[]::new);
        messages.add(
            switch (mCMessage.getRole()) {
              case "user" -> Message.user(new MessageContent(contents));
              case "system" -> Message.system(new MessageContent(contents));
```

</details>

**However,** I would assume extracting two methods `SingleChatMessage -> Message` and `MultiChatMessage -> Message` would result into even better code. This way you maybe can handle the duplicate code invoking `messages.add(Message)` that happens for both branches.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
Jonas-Isr: I am okay with deleting this :)

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
Jonas-Isr: > I would avoid two public APIs offering similar result.
I see that point! Will remove it :)

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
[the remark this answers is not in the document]
Jonas-Isr: Renamed to `withXyz()` as per vote in Slack :)

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
```diff
@@ -66,5 +99,5 @@ default ChatMessage createChatMessage() {
    */
   @Nonnull
   @Beta
-  String content();
+  MessageContent content();
```
newtork, line 102: So this is the expected API usage with your PR?

```java
OrchestrationChatResponse response = null;

for (Message msg : response.getAllMessages()) {
  for (ContentItem item : msg.content().contentItemList()) {
    if(item instanceof TextItem textItem) {
      System.out.println("Text item: "+textItem.text());
    }
    else if (item instanceof ImageItem imageItem) {
      System.out.println("Image item: "+imageItem.imageUrl());
    }
  }
}
```

I do like:
* `Message#content()` returns a complex object `MessageContent` for which we can later add more features if we like.

I do not like:
* `Message#content().contentItemList()` sounds repetitive. I would rather just name it `items()`. Then the usage would result in...

```diff
  Message msg;
  List<ContentItem> items = 
-   msg.content().contentItemList();
+   msg.content().items();
```

docs/guides/ORCHESTRATION_CHAT_COMPLETION.md
```diff
@@ -259,6 +259,51 @@ try (Stream<String> stream = client.streamChatCompletion(prompt, config)) {
...
+## Add images and multiple text inputs to a message
+
+It's possible to add images and multiple text inputs to a message.
+
+### Add images to a message
+
+An image can be added to an existing message as follows.
+
+```java
+var newMessage = message.andImage("https://url.to/image.jpg");
```
newtork, line 272: (Minor)

Please update the doc.

I want to implicitly (or explicitly) point out this only applies to `UserMessage` without mentioning this type directly.
Also "an existing message" may be to ambiguous for the reader.

```suggestion
var oldMessage = Message.user("Describe the following image");
var newMessage = message.withImage("https://url.to/image.jpg");
```

orchestration/src/main/java/com/sap/ai/sdk/orchestration/Message.java
Jonas-Isr: I like that idea! Added it :)

newtork (approved):

newtork (approved):

PR #297, opened 2025-01-23
fix: [OpenAI] Make system prompt message handling consistent

newtork:
We've noticed inconsistent handling of system prompt in convenience class `OpenAiClient`.

If the user defined a base system prompt, then it may be ignored on subsequent calls.
Instead of just logging a warning, we could harden the API contract(?)


rpanackal:
I like the solution. Though, I am a bit hesitant on including an additional interface for this. 

How are we feeling about removing the `withSystemPrompt()` feature?

newtork:
> How are we feeling about removing the `withSystemPrompt()` feature?

Not possible due to API stability.

PR #305, opened 2025-01-27
feat: [OpenAI] Stable Convenience API for OpenAI

rpanackal:
## Context

AI/ai-sdk-java-backlog#172.

We would like start using generated model classes for OpenAI module to keep up with spec changes easier.

Previously, thanks to Charles, we had a beautiful convenience API+model classes in place. But, we are migrating away from current convenience API to achieve parity with orchestration module. This equates to breaking changes.

100% convenience parity seems difficult due how both modules should treat ai model parameter configuration.

**Model Parameter Configuration** 

Under `orchestration` scenario, the ai model itself is selected on a per request level. While, in `foundation-model` scenario, we have a deployment specific to a model, while parameters alone (eg: temperature) are altered per request.

=> parameters scoping under a model, like in `OrchestrationAiModel` class, is not sensible in OpenAI convenience API.

### Feature scope:
 
- [x] Convenient request, response and error classes
- [ ] Convenient message classes (to match orchestration)
- [x] Code reduction in `OpenAiChatCompletionDelta`
- [x] Reduce num of mixins

## Definition of Done

- [ ] Documentation updated
- [ ] Release notes updated


core/src/test/java/com/sap/ai/sdk/core/AiCoreServiceTest.java
```diff
@@ -63,17 +68,30 @@ void testCustomBaseDestination() {
...
+    val server = new WireMockServer(wireMockConfig().dynamicPort());
+    server.start();
+    server.stubFor(get(urlEqualTo("/v2/lm/deployments")).willReturn(okJson(d)));
+    HttpDestination foobar =
+        service
+            .withBaseDestination(DefaultHttpDestination.builder(server.baseUrl() + "/v2/").build())
+            .getInferenceDestination()
+            .forScenario("foobar");
+    assertThat(foobar.getUri()).hasPath("/v2/inference/deployments/0123456789abcdef/");
+    server.stop();
```
newtork, line 94: (Question/Minor)

This looks like a drive-by(?). Or is it related to your PR? If not, I would rather move this out to a dedicated PR.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiAssistantMessage.java
```diff
@@ -0,0 +1,31 @@
...
+  /** The content of the message. */
+  @Nonnull String content;
+
+  /**
+   * Converts the message to a serializable object.
+   *
+   * @return the corresponding {@code ChatCompletionRequestAssistantMessage} object.
+   */
+  @Nonnull
+  public ChatCompletionRequestAssistantMessage createDTO() {
```
newtork, line 26: (Comment/Minor)

I understand you are mirroring the Orchestration API design, as suggested. However, I would like to challenge this method (visibility) here. We don't really want our convenience-users to access this method. While I can't challenge it in orchestration anymore (due to stability), maybe we can explore other options here in the new API. 

Can be done in a separate PR / ticket.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionOutput.java
```diff
@@ -0,0 +1,55 @@
...
+import com.sap.ai.sdk.foundationmodels.openai.model2.CreateChatCompletionResponseChoicesInner;
+import java.util.Objects;
+import javax.annotation.Nonnull;
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+
+/** Represents the output of an OpenAI chat completion. */
+@Value
+@RequiredArgsConstructor(access = PACKAGE)
+public class OpenAiChatCompletionOutput {
```
newtork, line 17: (Comment/Minor)

I was about to suggest `protected` as constructor visibility, such that users could still extend this class for testing/customization/workarounds. But then I realized we're using `@Value` annotation which renders the class `final`. Users will not be able to extend for testing/customization/workarounds.

Maybe I will suggest that in a separate PR, including _orchestration_.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionPrompt.java
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Creates an OpenAiChatCompletionPrompt with a single message.
+   *
+   * @param message the message to be added to the prompt
+   * @return a new OpenAiChatCompletionPrompt instance
+   */
+  @Nonnull
+  public static OpenAiChatCompletionPrompt create(@Nonnull final String message) {
+    return new OpenAiChatCompletionPrompt(message);
```
newtork, line 34: (Minor)

Please consider method chaining and minimal constructor:

```suggestion
@RequiredArgsConstructor(access = lombok.AccessLevel.PRIVATE)
public class OpenAiChatCompletionPrompt {
  @Nonnull private final List<OpenAiMessage> messages;

  @Nonnull private final List<String> stop = new ArrayList<>();

  /**
   * Creates an OpenAiChatCompletionPrompt with a single message.
   *
   * @param message the message to be added to the prompt
   * @return a new OpenAiChatCompletionPrompt instance
   */
  @Nonnull
  public static OpenAiChatCompletionPrompt create(@Nonnull final String message) {
    return create(OpenAiMessage.user(message));
```

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -110,65 +111,128 @@ public static OpenAiClient withCustomDestination(@Nonnull final Destination dest
...
-              .content(ChatCompletionRequestSystemMessageContent.create(systemPrompt)));
-    }
-    parameters
-        .addMessagesItem(
-            new ChatCompletionRequestUserMessage()
-                .role(ChatCompletionRequestUserMessage.RoleEnum.USER)
-                .content(ChatCompletionRequestUserMessageContent.create(prompt)))
-        .functions(null)
-        .tools(null)
-        .parallelToolCalls(null);
```
newtork, line 150: (Note)

This branch is setting `null` implicitly via passing `OpenAiChatCompletionConfig` fields, which are initialized as `null` in case of `List` value type.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -110,65 +111,128 @@ public static OpenAiClient withCustomDestination(@Nonnull final Destination dest
...
   }
 
+  private static void throwOnContentFilter(@Nonnull final OpenAiChatCompletionDelta delta) {
+    final String finishReason = delta.getFinishReason();
+    if (finishReason != null && finishReason.equals("content_filter")) {
+      throw new OpenAiClientException("Content filter filtered the output.");
+    }
+  }
+
+  private static CreateChatCompletionRequest toCreateChatCompletionRequest(
```
newtork, line 121: (Comment)

I appreciate this method remains private, in _orchestration_ its public unfortunately.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiError.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Value;
+import lombok.experimental.Delegate;
+
+/** Represents an error response from the OpenAI API. */
+@Beta
+@Value
+@AllArgsConstructor(onConstructor = @__({@JsonCreator}), access = AccessLevel.PROTECTED)
+public class OpenAiError implements ClientError {
```
newtork, line 17: (Comment)

I'm unhappy about the interface name "ClientError", since "Error" is reserved for non-recoverable exceptions leading to application shutdown.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiMessage.java
```diff
@@ -0,0 +1,49 @@
+package com.sap.ai.sdk.foundationmodels.openai;
+
+import com.sap.ai.sdk.foundationmodels.openai.model2.ChatCompletionRequestMessage;
+import javax.annotation.Nonnull;
+
+/** Interface representing convenience wrappers of chat message to the openai service. */
+public interface OpenAiMessage {
```
newtork, line 7: (Question)

Using this interface only as marker (without methods) is expected, yes? IIRC this is different in _orchestration_. Do we plan to include the image-improvements later, e.g. in a separate PR?

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionOutput.java
rpanackal: I have been noticing this problem of using `@Value` as well and has shifted to `@Data` in many places. We do expose complex type fields through getters and the internal state can be modified. I will apply this change.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiMessage.java
rpanackal: Yes. Currently, I didn't add complete convenience for messaging (eg: image, tools) and just matched with what is offered under orchestration.

Ideally, we should do this under a separate PR. Nonetheless, this would be a regression in terms of convenience offerings that was available in OpenAI module.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -110,65 +111,128 @@ public static OpenAiClient withCustomDestination(@Nonnull final Destination dest
...
-   * Generate a completion for the given prompt.
+   * Generate a completion for the given conversation.
    *
-   * @param parameters the prompt, including messages and other parameters.
+   * @param prompt The prompt, includes a list of messages from the conversation.
    * @return the completion output
+   * @throws OpenAiClientException
+   */
+  @Nonnull
+  public OpenAiChatCompletionOutput chatCompletion(@Nonnull final OpenAiChatCompletionPrompt prompt)
```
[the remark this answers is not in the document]
rpanackal: I felt the same way that there are too many chatCompletion methods and agree that we need to find a way to minimise this.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionOutput.java
rpanackal: For openai, I will include these change right way.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiEmbeddingsRequestFactory.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.List;
+import javax.annotation.Nonnull;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import lombok.experimental.Accessors;
+
+/** Factory class for creating OpenAI Embedding Requests. */
+@Accessors(fluent = true)
+@NoArgsConstructor(access = AccessLevel.PROTECTED)
+public class OpenAiEmbeddingsRequestFactory {
```
newtork, line 15: (Minor)

I don't think we'll support this public API in that shape:
* I think it should be `@Beta`
* "Factory" name is discouraged
* (Instead) rather have convenience class for embedding-create-request, that can be instantiated via _String(s)_. Then we wouldn't need this class.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/model2/ChatCompletionFunctionCallOption.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.Set;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Specifying a particular function via &#x60;{\&quot;name\&quot;: \&quot;my_function\&quot;}&#x60;
+ * forces the model to call that function.
+ */
+// CHECKSTYLE:OFF
+public class ChatCompletionFunctionCallOption
```
newtork, line 36: (Major)

We need to add `@Beta` to these classes. Manually I assume.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiEmbeddingsRequestFactory.java
rpanackal: Apart from the naming, can you expand a bit on why this is not a good approach?

My initial approach was what you suggested but later moved to the current state for the following reasons:

1. Avoid duplication of model class
2. User can call setters for generated model fields on top
3. Avoid adding an additional method in client (getting bloated) for accepting high level api object for embedding
4. In case, embedding become a bigger topic, we can add further static factory methods that returns a new convenience objects.

I might be missing issue that you are seeing.

rpanackal:
This PR has been split into two new ones
1. https://github.com/SAP/ai-sdk-java/pull/322
2. https://github.com/SAP/ai-sdk-java/pull/323

This conversation is now closed, but continued on the new PRs.

PR #323, opened 2025-02-06
feat: [OpenAI] Stable convenience - Latest 

rpanackal:
## Context

AI/ai-sdk-java-backlog#172.

**PR split out of https://github.com/SAP/ai-sdk-java/pull/305**

We would like start using generated model classes for OpenAI module to keep up with spec changes easier.

Previously, thanks to Charles, we had a beautiful convenience API+model classes in place. But, we are migrating away from current convenience API to achieve parity with orchestration module. This equates to breaking changes.

100% convenience parity seems difficult due how both modules should treat ai model parameter configuration.

**Model Parameter Configuration** 

Under `orchestration` scenario, the ai model itself is selected on a per request level. While, in `foundation-model` scenario, we have a deployment specific to a model, while parameters alone (eg: temperature) are altered per request.

=> parameters scoping under a model, like in `OrchestrationAiModel` class, is not sensible in OpenAI convenience API.

### Feature scope:
 
- [x] Add convenient request, response and error classes
- [x] Add convenient message classes (to match orchestration)
- [x] Add new `chatCompletion` and `streamChatCompletion` methods in client
- [x] Request class made immutable for config reuse and no getters  (`@lombok.Value` type)
- [x] Response class made immutable and no setters
- [x] `toCreateChatCompletionRequestMessage`, ie. toDTO equivalent, is non-public and moved to new utility class
- [x] Code reduction in `OpenAiChatCompletionDelta`
- [x] Reduce num of mixins
- [x] Adapt Tests
  - [x] Replicate unit and e2e for both new and old api 
    - [x] New service class introduced in test
      - Old service class still in demo
      - [x] Move common stubbing logic into a new parent test (consistent stubbing)
  - [x] Add test for tool call
  - [x] increase coverage for all api 

## Definition of Done

- [ ] Documentation updated
- [ ] Release notes updated


foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -44,6 +49,12 @@ public final class OpenAiClient {
 
   @Nonnull private final Destination destination;
 
+  static {
+    JACKSON.addMixIn(
```
newtork, line 53: (Minor)

I think we can move this invocation directly to the field initialization. I don't see direct benefit of having it here in a dedicated `static { }` block

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -119,12 +134,13 @@ public OpenAiClient withSystemPrompt(@Nonnull final String systemPrompt) {
...
 
   /**
-   * Generate a completion for the given user prompt.
+   * Generate a completion for the given string prompt as user.
    *
    * @param prompt a text message.
    * @return the completion output
    * @throws OpenAiClientException if the request fails
    */
+  @Deprecated(since = "1.3.0")
```
newtork, line 143: (Minor)

Good. But there should be a section in JavaDoc `@deprecated [...]` that explains what to use instead. This way developers have a chance to find the new recommended API.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -137,23 +153,54 @@ public OpenAiChatCompletionOutput chatCompletion(@Nonnull final String prompt)
...
-   * Generate a completion for the given prompt.
+   * Generate a completion for the given conversation and request parameters.
    *
-   * @param parameters the prompt, including messages and other parameters.
+   * @param request the completion request.
    * @return the completion output
    * @throws OpenAiClientException if the request fails
    */
   @Nonnull
+  public OpenAiChatCompletionResponse chatCompletion(
```
newtork, line 163: (Minor)

This is new API right? Please add `@since [...]` to the JavaDoc. Applies to all new public API.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -237,6 +344,35 @@ private void warnIfUnsupportedUsage() {
...
+  /**
+   * Get a vector representation of a given request with input that can be easily consumed by
+   * machine learning models and algorithms.
+   *
+   * @param request the request with input text.
+   * @return the embedding output
+   * @throws OpenAiClientException if the request fails
+   */
+  @Nonnull
+  public EmbeddingsCreate200Response embedding(@Nonnull final EmbeddingsCreateRequest request)
```
newtork, line 371: (Question)

Did you not deprecate the legacy `embedding(...)` method because we don't have convenience, yet? Or was this just missed?

rpanackal: It was just missed.

rpanackal: Additionally, would it be sensible to deprecate all old model classes? They have `@Beta` annotations on them.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/NewOpenAiClientTest.java
```diff
@@ -0,0 +1,647 @@
...
+import javax.annotation.Nonnull;
+import lombok.SneakyThrows;
+import org.assertj.core.api.SoftAssertions;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.Mockito;
+
+class NewOpenAiClientTest extends BaseOpenAiClientTest {
```
newtork, line 58: (Minor)

We cannot really call it "New", it's a relative term.
How about

```
BaseOpenAiClientTest -> OpenAiClientTest
OpenAiClientTest -> OpenAiClientTestLegacy
OpenAiClientTest -> OpenAiClientTestGenerated
```

Though I'm not fully convinced about the inheritance. It adds complexity to the unit test. Maybe there's a solution with composition. But we could also address that in a separate follow up.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java
```diff
@@ -24,8 +21,13 @@
...
 @SuppressWarnings("unused")
 public class OpenAiController {
-  @Autowired private OpenAiService service;
+  @Autowired private NewOpenAiService service;
 
+  /**
+   * Chat request to OpenAI
+   *
+   * @return a ResponseEntity with the response content
+   */
```
newtork, line 30: (Question)

Was this an intentional change? If yes, please ask Charles, why he deleted the JavaDocs previously.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
rpanackal: True. I kept it there so the field declarations look cleaner and no disadvantage in moving it to static block I believe.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java
```diff
@@ -46,14 +53,10 @@ ResponseEntity<ResponseBodyEmitter> streamChatCompletionDeltas() {
     final var emitter = new ResponseBodyEmitter();
     final Runnable consumeStream =
         () -> {
-          final var totalOutput = new OpenAiChatCompletionOutput();
           // try-with-resources ensures the stream is closed
           try (stream) {
-            stream
-                .peek(totalOutput::addDelta)
-                .forEach(delta -> send(emitter, delta.getDeltaContent()));
+            stream.forEach(delta -> send(emitter, delta.getDeltaContent()));
```
newtork, line 58: (Question)

Are you reducing the output to no longer print the usage stats? Just wondering, what's the rational?

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/NewOpenAiService.java
```diff
@@ -0,0 +1,190 @@
...
+                        ChatCompletionRequestUserMessageContent.create(
+                            List.of(
+                                new ChatCompletionRequestMessageContentPartText()
+                                    .text(message)
+                                    .type(TEXT)))))
+            .tools(null)
+            .functions(null)
+            .parallelToolCalls(null);
+
+    return OpenAiClient.forModel(GPT_35_TURBO).streamChatCompletionDeltas(request);
```
newtork, line 80: (Question)

Why are you not using the recommended convenience API?

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/NewOpenAiService.java
```diff
@@ -0,0 +1,190 @@
...
+            .role(USER)
+            .content(ChatCompletionRequestUserMessageContent.create(List.of(partText, partImage)));
+    final var request =
+        new CreateChatCompletionRequest()
+            .addMessagesItem(userMessage)
+            .functions(null)
+            .tools(null)
+            .parallelToolCalls(null);
+
+    return OpenAiClient.forModel(GPT_4O).chatCompletion(request);
```
newtork, line 124: (Comment)

I understand that we're waiting for orchestration multi-modal chat-message PR to be merged. Once available, we can update the openai convenience API. then this code here (based on generated classes) will hopefully not be necessary anymore.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/NewOpenAiService.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.Map;
+import java.util.stream.Stream;
+import javax.annotation.Nonnull;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+
+/** Service class for OpenAI service */
+@Service
+@Slf4j
+public class NewOpenAiService {
```
newtork, line 43: (Minor)

Please change class name.
We're using the sample application to showcase/demo our SDK to users. They will wonder about this name :)

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java
newtork: Compare here
https://github.com/SAP/ai-sdk-java/blob/4ebc2a34a00112e974961a13d525512f5cd3f09c/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java#L65-L77

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
newtork: > Additionally, would it be sensible to deprecate all old model classes? They have `@Beta` annotations on them.

I would leave the model classes `@Beta`. As long as the API methods are annotated `@Deprecated` the user is aware.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
newtork: > True. I kept it there so the field declarations look cleaner and no disadvantage in moving it to static block I believe.

I would agree if the field initialization would also happen in the same static code block.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -171,19 +220,21 @@ public OpenAiChatCompletionOutput chatCompletion(
...
    * @param prompt a text message.
-   * @return A stream of message deltas
+   * @return A stream of text chunks
    * @throws OpenAiClientException if the request fails or if the finish reason is content_filter
-   * @see #streamChatCompletionDeltas(OpenAiChatCompletionParameters)
+   * @see #streamChatCompletionDeltas(OpenAiChatCompletionRequest)
    */
   @Nonnull
   public Stream<String> streamChatCompletion(@Nonnull final String prompt)
       throws OpenAiClientException {
```
CharlesDuboisSAP, line 229: Why did you keep `streamChatCompletion(String)` with a behaviour change, but you deprecated `chatCompletion(String)` instead of doing the same behaviour change. It would be less breaking to also keep it,

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionResponse.java
```diff
@@ -0,0 +1,59 @@
...
+import lombok.RequiredArgsConstructor;
+
+/**
+ * Represents the output of an OpenAI chat completion. *
+ *
+ * @since 1.3.0
+ */
+@Data
+@RequiredArgsConstructor(access = PACKAGE)
+public class OpenAiChatCompletionResponse {
```
CharlesDuboisSAP, line 21: Naming this class `OpenAiChatCompletionOutput` would be one less breaking change.
Also I would like all these manually written model classes to be in their own package, this is one thing I regret we didn't do in orchestration

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/NewOpenAiService.java
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  /**
+   * Chat request to OpenAI with a tool.
+   *
+   * @param description of the function to be sent to the assistant
+   * @return the assistant message response
+   */
+  @Nonnull
+  public OpenAiChatCompletionResponse chatCompletionTools(@Nonnull final String description) {
```
CharlesDuboisSAP, line 122: Please remove the parameter, the description of the Fibonacci tool is fixed anyway

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
newtork: No behavior change
One method returns (stable) String, the other method returns deprecated class.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionResponse.java
newtork: > Naming this class `OpenAiChatCompletionOutput` would be one less breaking change.

It would still be a breaking change since namespace is different.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/NewOpenAiService.java
newtork: I agree.. the parameter by itself doesn't make too much sense.

We are looking at a "service" class here. That's a middle-layer between webapp backend and external services. We need to think about what could be a "realistic" API for that middle layer. Having no parameter would be counter-intuitive to me, for such a middle-layer.

Maybe we can change the service method here to consider the `int months` as parameter:
```diff
- How many rabbits will there be after 12 months")
+ How many rabbits will there be after %s months?".formatted(months))
```

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
CharlesDuboisSAP: I don't mind breaking changes.
`OpenAiClientTest.testThrowsOnContentFilter` was removed, I'm guessing this means `streamChatCompletion()` has a behaviour change.
If you had `chatCompletion(String)` not deprecated it could still exist but return the generated class.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionResponse.java
CharlesDuboisSAP: `wrapper`? also we still have those `public ChatCompletionRequestAssistantMessage createChatCompletionRequestMessage()`, we shouldn't have them public

CharlesDuboisSAP: Conclusion from the meeting:

It would be possible for us in the backend to create a ChatMessage without having public API

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/NewOpenAiService.java
rpanackal: This class has now been moved to test.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/NewOpenAiClientTest.java
rpanackal: `OpenAiClientTest -> OpenAiClientTestGenerated` is done

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionRequest.java
```diff
@@ -0,0 +1,262 @@
...
+            this.logprobs,
+            this.topLogprobs,
+            this.n,
+            parallelToolCalls,
+            this.seed,
+            this.streamOptions,
+            this.responseFormat,
+            this.tools,
+            this.toolChoice);
+  }
```
rpanackal, line 192: `@lombok.With` generates `withBooleanX` methods that does comparison of Boolean value with `==` operator . Unfortunately, this gets flagged as a bug (rightfully).

To keep nullability of variable,  we are forced to have these ugly methods in the class now.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
```diff
@@ -198,160 +70,96 @@ private static Callable<?>[] chatCompletionCalls() {
...
+                         "content" : "You are a helpful AI"
+                       }, {
+                         "role" : "user",
+                         "content" : [ {
+                           "type" : "text",
+                           "text" : "Hello World! Why is this phrase so famous?"
+                         } ]
+                       } ]
+                     }""")));
   }
```
CharlesDuboisSAP, line 159: The `history()` test is missing

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
```diff
@@ -380,48 +188,20 @@ void embedding() {
...
-    when(deltaWithContentFilter.getFinishReason()).thenReturn("content_filter");
-    when(mock.streamChatCompletionDeltas(any())).thenReturn(Stream.of(deltaWithContentFilter));
-
-    // this must not throw, since the stream is lazily evaluated
-    var stream = mock.streamChatCompletion("");
-    assertThatThrownBy(stream::toList)
-        .isInstanceOf(OpenAiClientException.class)
-        .hasMessageContaining("Content filter");
+                      {"input": ["Hello World"]}""")));
   }
```
CharlesDuboisSAP, line 193: The `testThrowsOnContentFilter()` is also missing.
Please compare both of your test classes `OpenAiClientTest` and `OpenAiClientGeneratedTest` to the `OpenAiClientTest` on the main branch to see what you missed

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
```diff
@@ -433,39 +213,28 @@ void streamChatCompletionDeltasErrorHandling() throws IOException {
...
+    try (var inputStream = stubStreamChatCompletion("streamChatCompletion.txt")) {
 
       final var request =
           new OpenAiChatCompletionParameters()
               .addMessages(
                   new OpenAiChatUserMessage()
                       .addText("Can you give me the first 100 numbers of the Fibonacci sequence?"));
 
-      try (Stream<OpenAiChatCompletionDelta> stream = client.streamChatCompletionDeltas(request)) {
+      try (Stream<com.sap.ai.sdk.foundationmodels.openai.model.OpenAiChatCompletionDelta> stream =
```
CharlesDuboisSAP, line 224:
```suggestion
      try (Stream<OpenAiChatCompletionDelta> stream =
```

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
```diff
@@ -510,9 +280,10 @@ void streamChatCompletionDeltas() throws IOException {
         assertThat(delta2.getObject()).isEqualTo("chat.completion.chunk");
         assertThat(delta2.getUsage()).isNull();
         assertThat(delta2.getPromptFilterResults()).isNull();
+        assertThat(delta2.getFinishReason()).isNull();
+
         final var choices2 = delta2.getChoices().get(0);
         assertThat(choices2.getIndex()).isEqualTo(0);
-        assertThat(choices2.getFinishReason()).isNull();
```
CharlesDuboisSAP, line 515: Please add this assertion back

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionResponse.java
```diff
@@ -0,0 +1,64 @@
...
+   */
+  @Nonnull
+  public String getContent() {
+    if (CONTENT_FILTER.equals(getOriginalResponse().getChoices().get(0).getFinishReason())) {
+      throw new OpenAiClientException("Content filter filtered the output.");
+    }
+
+    return Objects.requireNonNullElse(getChoice().getMessage().getContent(), "");
+  }
+}
```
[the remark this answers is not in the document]
CharlesDuboisSAP: You could have a `getText()` which returns String and a getContent which returns the `Content` class

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionResponse.java
CharlesDuboisSAP: Please remove the `createChatCompletionRequestMessage()` method @rpanackal

rpanackal: The method is now package private and perfectly enclosed the logic per message type. The mapping to low level object grows in the upcoming PR and I think especially its better to enclose the logic here.

Do you foresee a case where package-private visibility is problematic?

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionDelta.java
```diff
@@ -0,0 +1,78 @@
...
+import lombok.Setter;
+import lombok.Value;
+
+/**
+ * Represents an OpenAI chat completion output delta for streaming.
+ *
+ * @since 1.4.0
+ */
+@Beta
+@Value
```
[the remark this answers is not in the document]
rpanackal: Simply keeping both regular response and delta response object consistent. I didn't see a risk. 

(no strong opinion)

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientGeneratedTest.java
```diff
@@ -0,0 +1,571 @@
...
+        assertThat(deltaList).hasSize(5);
+        // the first two and the last delta don't have any content
+        assertThat(deltaList.get(0).getDeltaContent()).isEmpty();
+        assertThat(deltaList.get(1).getDeltaContent()).isEmpty();
+        assertThat(deltaList.get(2).getDeltaContent()).isEqualTo("Sure");
+        assertThat(deltaList.get(3).getDeltaContent()).isEqualTo("!");
+        assertThat(deltaList.get(4).getDeltaContent()).isEmpty();
+
+        assertThat(deltaList.get(3).getFinishReason()).isNull();
+        assertThat(deltaList.get(4).getFinishReason()).isEqualTo(STOP.getValue());
```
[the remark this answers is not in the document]
rpanackal: Call you explain how it is more convenient?

Currently, delta object may not contain choice, and even if it did it may be null. We handle the scenario for the user.

Additionally, we are implementing `StreamedDelta` which enforces a `String` return on `getFinishReason()`.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientGeneratedTest.java
```diff
@@ -0,0 +1,571 @@
...
+        client.embedding(
+            new EmbeddingsCreateRequest()
+                .input(EmbeddingsCreateRequestInput.create("Hello World")));
+
+    assertThat(result).isNotNull();
+    assertThat(result.getModel()).isEqualTo("ada");
+    assertThat(result.getObject()).isEqualTo("list");
+
+    assertThat(result.getUsage()).isNotNull();
+    assertThat(result.getUsage().getCustomFieldNames()).doesNotContain("completion_tokens");
```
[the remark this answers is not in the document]
rpanackal: I removed the suggested one alone.
```diff
-  assertThat(result.getUsage().getCustomFieldNames()).doesNotContain("completion_tokens");
```
@newtork added the rest. 

I believe this is to make sure delta objects are handled properly, especially as they have different properties present in different chunks by their chronology. This also acts as a documentation as the behaviour of incoming chunks may change in future (expected).

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientGeneratedTest.java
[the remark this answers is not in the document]
CharlesDuboisSAP: My bad, I thought I deleted this

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionDelta.java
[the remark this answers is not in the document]
CharlesDuboisSAP: It's not a risk it's just unnecessary, the customer will never create Response classes

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionDelta.java
```diff
@@ -0,0 +1,78 @@
...
+   * @return The completion usage or null.
+   */
+  @Nullable
+  public CompletionUsage getCompletionUsage(@Nonnull final ObjectMapper objectMapper) {
+    if (getOriginalResponse().getCustomFieldNames().contains("usage")
+        && getOriginalResponse().getCustomField("usage") instanceof Map<?, ?> usage) {
+      return objectMapper.convertValue(usage, CompletionUsage.class);
+    }
+    return null;
+  }
```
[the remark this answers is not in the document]
rpanackal: @newtork do you remember why? [original pull request](https://github.com/SAP/ai-sdk-java/blob/4ebc2a34a00112e974961a13d525512f5cd3f09c/foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionDelta.java#L80)

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
rpanackal: I was assuming this need not be tested twice as this feature wouldn't change across two apis.

But I guess, the usecase of withSystemPrompt() + chatCompletion(String) is more relevant under old api. So, I have replicated the test.

Let me know if you still think this test is needed for both api.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
rpanackal: This test is no longer relevant for old client methods (hence on OpenAiClientTest), as the streamChatCompletion(String) internally uses  the new convenience api.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
CharlesDuboisSAP: One API is good enough

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
rpanackal: It is being tested  indirectly through the high level api call.

https://github.com/SAP/ai-sdk-java/blob/d4a03e5dc36844bfedaa0f16d6f825e7fde8ffd7/foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java#L283

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionDelta.java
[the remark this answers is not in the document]
newtork: When the code was written there was no general available (static?) `ObjectMapper` instance. Please feel free to replace/remove the method argument. As long as the tests pass.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientGeneratedTest.java
```diff
@@ -0,0 +1,571 @@
...
+        assertThat(delta4Choice.getDelta().getToolCalls()).isEmpty();
+        assertThat(delta4.getChoices()).hasSize(1);
+        final var choice = delta4.getChoices().get(0);
+        assertThat(choice.getFinishReason()).isEqualTo(STOP);
+        final Map<?, ?> filterRaw = (Map<?, ?>) choice.getCustomField("content_filter_results");
+        assertThat(filterRaw).isEmpty();
+        assertThat(choice.getDelta()).isNotNull();
+        assertThat(choice.getDelta().getRole()).isNull();
+        assertThat(choice.getDelta().getContent()).isNull();
+        assertThat(choice.getDelta().getToolCalls()).isNotNull().isEmpty();
```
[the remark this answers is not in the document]
newtork: I'm not sure, whether JUnit will make a proper null check, or it'll throw a `NullPointerException` when attempting to compare the String 🤔

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionRequest.java
newtork: Can you explain the problem?

```java
Boolean true1 = true;
Boolean true2 = Boolean.TRUE;
Boolean true3 = Boolean.parseBoolean("true");

System.out.println(true1 == true2); // true
System.out.println(true1 == true3); // true
System.out.println(true2 == true3); // true
```

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionRequest.java
```diff
@@ -99,54 +110,151 @@ public OpenAiChatCompletionRequest(@Nonnull final String message) {
    * @param message the primary message to be added to the prompt
    * @param messages additional messages to be added to the prompt
    */
+  @Tolerate
   public OpenAiChatCompletionRequest(
       @Nonnull final OpenAiMessage message, @Nonnull final OpenAiMessage... messages) {
+    // Keeps default values for boolean fields. @With introduces bug comparison of Boolean
+    this(
+        new ArrayList<OpenAiMessage>(),
```
newtork, line 118: (Minor)

Please use Guava for convenient + immutable list creation.

```suggestion
        Lists.asList(message, messages),,
```

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionRequest.java
rpanackal: The issue is with the (now deprecated) way of creating Boolean instances like `new Boolean(true)`. I am assuming the bug warning is still valid.

```
Boolean b1 = new Boolean(true);
Boolean b2 = new Boolean(true);
System.out.println(b1 == b2);
```

CharlesDuboisSAP: It's about b1 == b2, doing this can easily be a bug, but in this case it's not

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionRequest.java
```diff
@@ -99,54 +110,151 @@ public OpenAiChatCompletionRequest(@Nonnull final String message) {
...
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        null);
```
newtork, line 136: (Comment)

I was investigating whether `@NoArgsConstructor` would work. And yes it does, but unfortunately we would then need to (1) make `messages` a mutable list and (2) write a _wither_ for it manually, which would invoke all-args constructor again. Therefore no code code would be saved.

On a second note: In the future, we could add `DEFAULT_XYZ` here instead of `null`. They would be default values for the model fields - currently all of them null / undefined anyway. But it could be useful in the future, and for documentation purpose. We may consider this in future PRs, not part of your branch.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientGeneratedTest.java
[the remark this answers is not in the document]
CharlesDuboisSAP: It will say expecting `null` to be `""` I'm guessing

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientGeneratedTest.java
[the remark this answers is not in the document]
CharlesDuboisSAP: No it was because we the custom field was previously used, now that the spec has increased, it no longer goes in the custom field and the assertion can be removed (or replaced with `getCompletionToken()` in this case)

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionRequest.java
rpanackal: I considered including the default but 
1. adding default values means the withX will not be generated.
2. also, keeping track of defaults and as it may change for the model objects
   - could load from the generated model object itself - but maybe not so neat
3. Say, we set it to `true`, it would now be part of all serialised json values and does scale well for every property and message frequency.

I don't believe any of them are dealbreakers but still hesitant.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientGeneratedTest.java
```diff
@@ -0,0 +1,575 @@
...
+        assertThat(delta0.getSystemFingerprint()).isNull();
+        assertThat(delta1.getSystemFingerprint()).isEqualTo("fp_e49e4201a9");
+        assertThat(delta2.getSystemFingerprint()).isEqualTo("fp_e49e4201a9");
+        assertThat(delta3.getSystemFingerprint()).isEqualTo("fp_e49e4201a9");
+        assertThat(delta4.getSystemFingerprint()).isEqualTo("fp_e49e4201a9");
+
+        assertThatThrownBy(() -> delta0.getCustomField("usage"))
+            .isInstanceOf(NoSuchElementException.class);
+        assertThat(delta1.getCustomField("usage")).isNull();
+        assertThat(delta1.getCustomField("usage")).isEqualTo(deltaList.get(1).getCompletionUsage());
```
CharlesDuboisSAP, line 380: I still don't understand what you are testing, it should be `getter`==`static data`.
You should probably get rid of `getCustomField("usage")` entirely.
You should assign `getCompletionUsage()` to a variable and assert on every getter of `CompletionUsage`

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientGeneratedTest.java
```diff
@@ -0,0 +1,575 @@
...
+        assertThat(delta2.getCustomField("usage")).isNull();
+        assertThat(delta3.getCustomField("usage")).isNull();
+        final Map<?, ?> delta4UsageRaw = (Map<?, ?>) delta4.getCustomField("usage");
+        final var delta4Usage = MAPPER.convertValue(delta4UsageRaw, CompletionUsage.class);
```
CharlesDuboisSAP, line 384: replace this with the proper getter

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiAssistantMessage.java
```diff
@@ -0,0 +1,37 @@
...
+  /** The content of the message. */
+  @Nonnull String content;
+
+  /**
+   * Converts the message to a serializable object.
+   *
+   * @return the corresponding {@code ChatCompletionRequestAssistantMessage} object.
+   */
+  @Nonnull
+  ChatCompletionRequestAssistantMessage createChatCompletionRequestMessage() {
```
CharlesDuboisSAP, line 32: We can get rid of this if we use a translator class like `ConfigToRequestTransformer`

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionDelta.java
```diff
@@ -0,0 +1,79 @@
...
+@EqualsAndHashCode
+public class OpenAiChatCompletionDelta implements StreamedDelta {
+  /** The original response from the chat completion stream. */
+  @Nonnull private final CreateChatCompletionStreamResponse originalResponse;
+
+  /**
+   * Retrieves the delta content from the original response.
+   *
+   * @return The delta content as a string, or an empty string if not available.
+   */
```
CharlesDuboisSAP, line 38:
```suggestion
```
@Override takes the Javadoc from the parent class

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionDelta.java
```diff
@@ -0,0 +1,79 @@
...
+      return Objects.requireNonNullElse(message.getContent(), "");
+    }
+    return "";
+  }
+
+  /**
+   * Retrieves the finish reason from the original response.
+   *
+   * @return The finish reason as a string, or null if not available.
+   */
```
CharlesDuboisSAP, line 54:
```suggestion
```

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionRequest.java
```diff
@@ -0,0 +1,252 @@
...
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+@Getter(value = AccessLevel.NONE)
+public class OpenAiChatCompletionRequest {
+  /** List of messages from the conversation. */
+  @Nonnull List<OpenAiMessage> messages;
+
+  /** Stop sequences for the completion. */
+  @Nullable List<String> stop;
+
+  /** Temperature for the completion. */
```
CharlesDuboisSAP, line 41: Can we have better Javadoc in here?
An example of stop sequences, what they do.
What number we can put in for temperature?
What is the diff between `maxTokens` and `maxCompletionTokens`?
etc

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionRequest.java
```diff
@@ -0,0 +1,252 @@
...
+  @Nullable
+  Boolean logprobs;
+
+  /** Number of top log probabilities to include. */
+  @Nullable Integer topLogprobs;
+
+  /** Number of completions to generate. */
+  @Nullable Integer n;
+
+  /** Whether to allow parallel tool calls. */
```
CharlesDuboisSAP, line 76: great Javadoc 👍

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClient.java
```diff
@@ -107,7 +111,11 @@ public static OpenAiClient withCustomDestination(@Nonnull final Destination dest
   }
 
   /**
-   * Add a system prompt before user prompts.
+   * Use this method to set a system prompt that should be used across multiple chat completions
+   * with basic string prompts {@link #streamChatCompletionDeltas(OpenAiChatCompletionRequest)}.
+   *
+   * <p>Note: The system prompt is ignored on chat completions invoked with
+   * OpenAiChatCompletionPrompt.
```
CharlesDuboisSAP, line 118: Did you mean `OpenAiChatCompletionRequest`?
Also if this method only works for the deprecated chatCompletion, then it should be deprecated as well

rpanackal: Missed the update in javadoc. Thanks for the catch.

We have room to choose there. The `withSystemPrompt` is effective both `chatCompletion(String)` and `streamChatCompletion(String)`. We may choose to keep it or remove it based on deprecation plan in future.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientGeneratedTest.java
rpanackal: I have split the test (low level | convenience) and re-organized it.

foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientGeneratedTest.java
rpanackal: Refer to https://github.com/SAP/ai-sdk-java/pull/323#discussion_r1961236542

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionDelta.java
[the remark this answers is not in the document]
rpanackal: I have now introduced a utility method to fetch and config an object mapper. The user no longer have to pass the object mapper. No utility methods are public.

CharlesDuboisSAP (approved):

PR #324, opened 2025-02-06
chore: DevOps `mvn install` fail early

CharlesDuboisSAP:
## Context

`mvn install` should fail as fast as possible regardless of what step fails.

### Feature scope:
 
- [x] Organized plugins by phase (cosmetic)
- [x] Moved SpotBugs to run after Jacoco (it never fails and takes a while, it is now in last)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -540,7 +540,14 @@
           </execution>
         </executions>
       </plugin>
-      <!-- phase: process-test-classes -->
+      <!-- phase: test -->
+      <plugin>
+        <!-- Needed because GitHub actions uses an old Maven version <3.9.0 -->
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>${surefire.version}</version>
+      </plugin>
+      <!-- phase: verify -->
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-checkstyle-plugin</artifactId>
@@ -566,7 +573,7 @@
             <goals>
               <goal>check</goal>
             </goals>
-            <phase>process-test-classes</phase>
+            <phase>verify</phase>
           </execution>
         </executions>
       </plugin>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -584,19 +627,21 @@ https://gitbox.apache.org/repos/asf?p=maven-pmd-plugin.git;a=blob_plain;f=src/ma
...
             </goals>
+            <phase>initialize</phase>
           </execution>
           <execution>
             <id>default-report</id>
             <goals>
               <goal>report</goal>
             </goals>
-            <phase>prepare-package</phase>
+            <phase>verify</phase>
```
CharlesDuboisSAP, line 637: Jacoco is after CheckStyle and PMD

MatKuhr (changes requested):
This is not suitable when doing PoCs or larger refactorings. There we don't want to fix all PMD and Javadoc requirements just to compile.

Instead, if you want to run all checks fast (i.e. before/without tests), just run `mvn install -DskipTests` or create an alias for that. Compile takes around 10s (on my machine), so there is no point in moving checks to before that for speed

CharlesDuboisSAP:
> This is not suitable when doing PoCs or larger refactorings. There we don't want to fix all PMD and Javadoc requirements just to compile.

CheckStyle and PMD still run after compile.

MatKuhr:
Okay fair, but still it's beneficial to run tests without fixing all javadoc warnings, why not just `mvn install -DskipTests`? our CI/CD pipeline does that already anyway..

CharlesDuboisSAP:
> Okay fair, but still it's beneficial to run tests without fixing all javadoc warnings, why not just `mvn install -DskipTests`? our CI/CD pipeline does that already anyway..

I run tests with IntelliJ, the only time I run `mvn install` is to make sure my PR is green

pom.xml
```diff
@@ -584,19 +626,21 @@ https://gitbox.apache.org/repos/asf?p=maven-pmd-plugin.git;a=blob_plain;f=src/ma
             <goals>
               <goal>prepare-agent</goal>
             </goals>
+            <phase>initialize</phase>
```
MatKuhr, line 629: what does this change do?

MatKuhr (approved):
lgtm

pom.xml
CharlesDuboisSAP: It's redundant, the default is already initialize

PR #330, opened 2025-02-13
feat: [OpenAI]  Tool Choice Convenience

rpanackal:
## Context

AI/ai-sdk-java-backlog#172.

Convenience request class contains fields of generated model types. Ideally, we want to limit scenarios the user would have to deal with generated model classes.

Tool choice convenience is lets the user to set the strategy for tools usage/selection.

### Feature scope:
 
- [x] Convenience methods for tool use in request construction

## Definition of Done

- [ ] Documentation updated
- [ ] Release notes updated


foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionRequest.java
```diff
@@ -126,7 +128,9 @@ public class OpenAiChatCompletionRequest {
   @Nullable List<ChatCompletionTool> tools;
 
   /** Option to control which tool is invoked by the model. */
-  @Nullable ChatCompletionToolChoiceOption toolChoice;
+  @With(AccessLevel.PRIVATE)
```
CharlesDuboisSAP, line 131: This is a breaking change, It's fine because we didn't release but be careful with those lombok annotations, they can create too much public API

CharlesDuboisSAP (approved):

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiChatCompletionRequest.java
```diff
@@ -248,6 +252,51 @@ public OpenAiChatCompletionRequest withLogprobs(@Nonnull final Boolean logprobs)
             this.toolChoice);
   }
 
+  /**
+   * Only message generation will be performed without calling any tool.
+   *
+   * @return the current OpenAiChatCompletionRequest instance.
+   */
+  @Nonnull
+  public OpenAiChatCompletionRequest withToolChoiceNone() {
```
newtork, line 261: (Comment)

Not a huge fan of having these duplicate methods signatures, only differing by an apparent argument - that actually is not a real method argument. Apart from the `String` based variant.

PR #341, opened 2025-02-19
feat: [Orchestration] Convenience for response format

Jonas-Isr:
## Context

AI/ai-sdk-java-backlog#195.

This PR adds convenience methods to set the response format in Orchestration. A JSON schema can be created from a class and then used as a response schema as follows:
```
var schema =
        ResponseJsonSchema.from(Translation.class)
            .withDescription("Output schema for language translation.")
            .withStrict(true);
val configWithResponseSchema = config.withTemplateConfig(TemplateConfig.create().withJsonSchemaResponse(schema));
```
If only any JSON response is wanted, without strict adherence to a given schema, the user can use 
`var configWithJsonResponse = config.withTemplateConfig(TemplateConfig.create().withJsonResponse())`.

### Feature scope:
 
- [x] Add a way to generate schema conveniently
    - [x] Add new class `ResponseJsonSchema`
    - [x] Add methods to create such schema from class or from a map
- [x] Add convenience layer for templates
    - [x] Add new classes `TemplateConfig`, `OrchestrationTemplate`, and `OrchestrationTemplateReference`
    - [x] convenience methods to add response format (JSON schema or JSON object) to config via `TemplateConfig`

## Definition of Done

- [x] Documentation updated
- [x] Release notes updated


docs/guides/ORCHESTRATION_CHAT_COMPLETION.md
```diff
@@ -334,38 +326,17 @@ Note, that it is necessary to tell the AI model to actually return a JSON object
...
-var template = Message.user("Whats '%s' in German?".formatted(word));
 var schema =
-        Map.of(
-                "type",
-                "object",
-                "properties",
-                Map.of(
-                        "language", Map.of("type", "string"),
-                        "translation", Map.of("type", "string")),
-                "required",
```
a-d, line 346: (Minor)

I would prefer if we kept the low level documented somewhere? (less prominent)
We found developers sometimes ask for low level stuff when having edge cases.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationModuleConfig.java
```diff
@@ -211,4 +220,95 @@ public OrchestrationModuleConfig withGrounding(
...
+   * @since 1.4.0
+   */
+  @Nonnull
+  public OrchestrationModuleConfig withJsonResponse() {
+    val responseFormatJsonObject =
+        ResponseFormatJsonObject.create().type(ResponseFormatJsonObject.TypeEnum.JSON_OBJECT);
+    if (this.templateConfig instanceof Template template) {
+      template.setResponseFormat(responseFormatJsonObject);
+      return this.withTemplateConfig(template);
+    }
```
newtork, line 309: (Question/Minor)

This has side-effect doesn't it?

The following code is (non) functional equivalent:
```suggestion
    if (this.templateConfig instanceof Template template) {
      template.setResponseFormat(responseFormatJsonObject);
      return this;
    }
```

Resulting (unexpectedly) in...
```java
val gpt4oCustomInstance = new OrchestrationModuleConfig().withLlmConfig(GPT_4O_MINI);
val configWithJsonResponse = gpt4oCustomInstance.withJsonResponse();

gpt4oCustomInstance.equals(configWithJsonResponse) // true
```

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationModuleConfig.java
```diff
@@ -211,4 +220,95 @@ public OrchestrationModuleConfig withGrounding(
...
+    if (this.templateConfig instanceof Template oldTemplate) {
+      responseFormat = oldTemplate.getResponseFormat();
+    }
+    //  Template.getResponseFormat() might return null, so the following check is necessary.
+    if (newTemplate != null && newTemplate.getResponseFormat() == null) {
+      newTemplate.setResponseFormat(responseFormat);
+    }
+    if (newTemplate == null && responseFormat != null) {
+      newTemplate = Template.create().template(List.of());
+      newTemplate.setResponseFormat(responseFormat);
```
newtork, line 255: (Question)
Why not..?
```suggestion
      newTemplate.setResponseFormat(null);
```

Does the request behavior differently?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationModuleConfig.java
```diff
@@ -211,4 +220,95 @@ public OrchestrationModuleConfig withGrounding(
...
+   * @param templateConfig The template configuration to use.
+   * @return A new configuration with the given template configuration.
+   */
+  @Tolerate
+  @Nonnull
+  public OrchestrationModuleConfig withTemplateConfig(
+      @Nullable final TemplatingModuleConfig templateConfig) {
+
+    //    If new templateConfig is a TemplateRef, use it.
+    if (templateConfig instanceof TemplateRef) {
```
newtork, line 238: (Minor/Question)

Maybe the code could be easier if you didn't have the interface as argument, but the concrete sub-classes instead. As a side effect we wouldn't struggle with more generated sub-classes in the future, where the behavior would be unexpected.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationModuleConfig.java
```diff
@@ -211,4 +220,95 @@ public OrchestrationModuleConfig withGrounding(
...
+
+  /**
+   * Creates a new configuration with the given template configuration.
+   *
+   * @link <a href="https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/templating">SAP
+   *     AI Core: Orchestration - Templating</a>
+   * @param templateConfig The template configuration to use.
+   * @return A new configuration with the given template configuration.
+   */
+  @Tolerate
```
newtork, line 232: (Question)
Why do you need Tolerate?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationModuleConfig.java
Jonas-Isr: `null` is allowed right now to be backward compatible.

docs/guides/ORCHESTRATION_CHAT_COMPLETION.md
Jonas-Isr: I put it in a collapsable. This also enables us to show the second convenience function.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationModuleConfig.java
Jonas-Isr: Tried to refactor the code to make it more readible.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationModuleConfig.java
Jonas-Isr: As discussed, this is not possible since we have to allow `null` as input for backward compatibility.

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/OrchestrationTest.java
```diff
@@ -301,8 +301,6 @@ void testResponseFormatJsonSchema() {
     final var result = service.responseFormatJsonSchema("apple").getOriginalResponse();
     final var choices = ((LLMModuleResultSynchronous) result.getOrchestrationResult()).getChoices();
     assertThat(choices.get(0).getMessage().getContent()).isNotEmpty();
-    assertThat(choices.get(0).getMessage().getContent()).contains("\"language\":\"German\"");
-    assertThat(choices.get(0).getMessage().getContent()).contains("\"translation\":\"Apfel\"");
```
CharlesDuboisSAP, line 305: Why?

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OrchestrationService.java
```diff
@@ -350,40 +350,29 @@ public OrchestrationChatResponse grounding(@Nonnull final String userMessage) {
    */
   @Nonnull
   public OrchestrationChatResponse responseFormatJsonSchema(@Nonnull final String word) {
-    final var llmWithImageSupportConfig =
-        new OrchestrationModuleConfig().withLlmConfig(GPT_4O_MINI);
+    val gpt4oCustomInstance = new OrchestrationModuleConfig().withLlmConfig(GPT_4O_MINI);
```
CharlesDuboisSAP, line 353: Optional: We could use `GPT_4O_MINI` everywhere.
Also what is custom about it?

orchestration/src/test/java/com/sap/ai/sdk/orchestration/OrchestrationModuleConfigTest.java
```diff
@@ -173,4 +197,101 @@ void testGroundingPrompt() {
...
+    assertThat(((Template) config.getTemplateConfig())).isNotNull();
+    assertThat(
+            ((ResponseFormatJsonSchema) ((Template) config.getTemplateConfig()).getResponseFormat())
+                .getJsonSchema()
+                .getSchema())
+        .isEqualTo(schema.getSchemaMap());
+  }
+
+  @Test
+  void testResponseFormatOverwrittenByNewTemplateConfigWithResponseFormat() {
```
CharlesDuboisSAP, line 234: Should the overwriting behaviours throw?
It could avoid customers asking why their response format is broken

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationModuleConfig.java
```diff
@@ -211,4 +220,118 @@ public OrchestrationModuleConfig withGrounding(
...
+    return new OrchestrationModuleConfig(
+        llmConfig,
+        ResponseJsonSchema.newTemplateWithResponseFormat(newTemplate, responseFormat),
+        maskingConfig,
+        filteringConfig,
+        groundingConfig);
+  }
+
+  @Nullable
+  private static TemplateResponseFormat getResponseFormatIfExists(
```
CharlesDuboisSAP, line 275:
```suggestion
  private static TemplateResponseFormat getResponseFormat(
```

CharlesDuboisSAP (commented):
Minor comments

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OrchestrationService.java
Jonas-Isr: I called it "custom" because it is different from the AI we use for most of the other tests. I think we went away from using GPT in all tests to not hit any usage quotas at one point? But I might be mistaking there.

orchestration/src/test/java/com/sap/ai/sdk/orchestration/OrchestrationModuleConfigTest.java
Jonas-Isr: I think if the users set a response format in their new template config they will expect it to change, right? So I don't think we should throw here.

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/OrchestrationTest.java
Jonas-Isr: Changed to have some tests again.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OrchestrationService.java
Jonas-Isr: Renamed it.

CharlesDuboisSAP (dismissed):

docs/guides/ORCHESTRATION_CHAT_COMPLETION.md
```diff
@@ -330,14 +332,42 @@ var schema =
...
+                List.of("language", "translation"),
+                "additionalProperties",
+                false);
+var schemaFromMap = ResponseJsonSchema.of(schemaMap, "Translator-Schema");
+var config = new OrchestrationModuleConfig()
+        .withLlmConfig(OrchestrationAiModel.GPT_4O);
+var configWithResponseSchema = config.withJsonSchemaResponse(schemaFromMap);
+
+var prompt = new OrchestrationPrompt(Message.user("Some message."));
+var response = client.chatCompletion(prompt, configWithTemplate).getContent();
```
newtork, line 366: A little formatting

```suggestion
var schemaMap =
    Map.ofEntries(
        entry("type", "object"),
        entry("properties", Map.ofEntries(
            entry("language", Map.of("type", "string")),
            entry("translation", Map.of("type", "string"))),
        entry("required", List.of("language","translation")),
        entry("additionalProperties", false)));

var schemaFromMap = ResponseJsonSchema.of(schemaMap, "Translator-Schema");
var config = new OrchestrationModuleConfig()
    .withLlmConfig(OrchestrationAiModel.GPT_4O);
var configWithResponseSchema = config.withJsonSchemaResponse(schemaFromMap);
```

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationModuleConfig.java
```diff
@@ -211,4 +212,21 @@ public OrchestrationModuleConfig withGrounding(
...
+   *
+   * @link <a href="https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/templating">SAP
+   *     AI Core: Orchestration - Templating</a>
+   * @param templateConfig The template configuration to use.
+   * @return A new configuration with the given template configuration.
+   * @since 1.5.0
+   */
+  @Tolerate
+  @Nonnull
+  @Beta
```
Jonas-Isr, line 227: I added `@Beta` annotations to every new convenience method/class.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationModuleConfig.java
```diff
@@ -211,4 +212,21 @@ public OrchestrationModuleConfig withGrounding(
...
   }
+
+  /**
+   * Creates a new configuration with the given template configuration as {@link TemplateConfig}.
+   *
+   * @link <a href="https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/templating">SAP
+   *     AI Core: Orchestration - Templating</a>
+   * @param templateConfig The template configuration to use.
+   * @return A new configuration with the given template configuration.
+   * @since 1.5.0
```
Jonas-Isr, line 223: I added `@since 1.5.0` everywhere. 

If we merge this before the 1.4.0 release, this has to be adapted.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationTemplateReference.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * A reference to a template to use in {@link OrchestrationModuleConfig}.
+ *
+ * @since 1.5.0
+ */
+@Value
+@AllArgsConstructor(access = AccessLevel.PROTECTED)
+@Beta
+public class OrchestrationTemplateReference extends TemplateConfig {
```
Jonas-Isr, line 20: I did not add tests for this class in `OrchestrationUnitTests` since this is part of [BLI 184](https://github.com/SAP/ai-sdk-java-backlog/issues/184)

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationTemplate.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+@EqualsAndHashCode(callSuper = true)
+@Value
+@With
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+@NoArgsConstructor(force = true, access = AccessLevel.PACKAGE)
+@Beta
+public class OrchestrationTemplate extends TemplateConfig {
+  @Nullable List<ChatMessage> template;
+  @Nonnull Map<String, String> defaults = new HashMap<>();
```
newtork, line 40: (Minor/Question)

What is the expected usage for `defaults`? Only exposed api is `getDefaults()`, is the expectation to have user run `put` on this object?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationTemplate.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @With(AccessLevel.PRIVATE)
+  @Nullable
+  TemplateResponseFormat responseFormat;
+
+  @Nonnull List<ChatCompletionTool> tools = new ArrayList<>();
+  @Nonnull Map<String, Object> cloudSdkCustomFields = new LinkedHashMap<>();
```
newtork, line 47: (Minor) (Contd.

Also here are collections that are accessible via `get[...]()` - is this expected?

image

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ResponseJsonSchema.java
```diff
@@ -0,0 +1,88 @@
...
+        new SchemaGenerator(
+            new SchemaGeneratorConfigBuilder(SchemaVersion.DRAFT_2020_12, OptionPreset.PLAIN_JSON)
+                .without(Option.SCHEMA_VERSION_INDICATOR)
+                .with(Option.FORBIDDEN_ADDITIONAL_PROPERTIES_BY_DEFAULT)
+                .with(module)
+                .build());
+    val jsonSchema = generator.generateSchema(classType);
+    val mapper = new ObjectMapper();
+    final Map<String, Object> schemaMap = mapper.convertValue(jsonSchema, new TypeReference<>() {});
+    val schemaName = ((Class<?>) classType).getSimpleName() + "-Schema";
```
newtork, line 85: (Question)

The `-` is fine for a name, right?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ResponseJsonSchema.java
```diff
@@ -0,0 +1,88 @@
...
+    val generator =
+        new SchemaGenerator(
+            new SchemaGeneratorConfigBuilder(SchemaVersion.DRAFT_2020_12, OptionPreset.PLAIN_JSON)
+                .without(Option.SCHEMA_VERSION_INDICATOR)
+                .with(Option.FORBIDDEN_ADDITIONAL_PROPERTIES_BY_DEFAULT)
+                .with(module)
+                .build());
+    val jsonSchema = generator.generateSchema(classType);
+    val mapper = new ObjectMapper();
+    final Map<String, Object> schemaMap = mapper.convertValue(jsonSchema, new TypeReference<>() {});
```
newtork, line 84: (Comment)

In case this haunted you ;)

```suggestion
    val schemaMap = mapper.convertValue(jsonSchema, new TypeReference<Map<String,Object>>() {});
```

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationTemplate.java
Jonas-Isr: I missed that the `.with...` were not generated here. 

I changed the code to do that and also added a test for this in `OrchestrationConvenienceUnitTest`.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationTemplate.java
Jonas-Isr: Removed `CloudSdkCustomFields` (that was a copy-paste mistake 😬)

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ResponseJsonSchema.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /**
+   * Create a new instance of {@link ResponseJsonSchema} from a given class.
+   *
+   * @param classType The class to generate the schema from
+   * @return The new instance of {@link ResponseJsonSchema}
+   */
+  @Nonnull
+  public static ResponseJsonSchema from(@Nonnull final Type classType) {
```
newtork, line 71: Is there a reason to use `from` and `of` for method names? I don't immediately notice a difference in semantics. Maybe we could align them?

Additionally I'm wondering whether we should put additional qualifiers to the name
```diff
- ResponseJsonSchema.of( ... )
+ ResponseJsonSchema,ofMap( ... ) 
+ ResponseJsonSchema.fromMap( ... ) // alternative


- ResponseJsonSchema.from( ... )
+ ResponseJsonSchema.fromType( ... )
+ ResponseJsonSchema.ofType( ... ) // alternative
```

Jonas-Isr: I tried to stick closely to the naming conventions in the __Effective Java__ book:
> from—A type-conversion method that takes a single parameter and returns a
corresponding instance of this type
of—An aggregation method that takes multiple parameters and returns an instance of this type that incorporates them

But I totally get that I might have been overdoing it there and that aligning the names has more value. I also like the additional qualifiers.

Jonas-Isr: I think I like `fromMap()` and `fromType()` best.

docs/release-notes/release_notes.md
```diff
@@ -15,7 +15,9 @@
 
 ### ✨ New Functionality
 
-- [Orchestration] [Add Spring AI tool calling](../guides/SPRING_AI_INTEGRATION.md#tool-calling).
+- [Orchestration]
```
newtork, line 18: (Comment)

We'll have to change / define a rule for grouping by module. Using `[Foo]` is neither valid markdown, nor improved readability. May not be addressed in this PR.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ResponseJsonSchema.java
newtork: > of—An aggregation method that takes multiple parameters and returns an instance of this type that incorporates them.

This got me confused at first, but I found the authors example:

```java
Set<Rank> faceCards = EnumSet.of(JACK, QUEEN, KING);
```

And it definitely checks for `List.of(T...)`,  `Set.of(T...)`,  `Map.of(T...)`.
But it seems like it only works best for enumerations.

I didn't find an explicit paragraph that could motivate `fromMap(...)` and `fromType(...)` apart from pointing out the advantage over constructors to have explanatory names:

> If the parameters to a constructor do not, in and of themselves, describe the object being returned, a static factory with a well-chosen name is easier to use and the resulting client code easier to read.

docs/release-notes/release_notes.md
Jonas-Isr: I opened a vote in Slack so that we can decide on this and then open a specific PR for our solution :)

newtork (approved):
LGTM

PR #348, opened 2025-02-24
fix: Update model lists and model availability test

TillK17:
## Context

- fix for [failed e2e-test](https://github.com/SAP/ai-sdk-java/actions/runs/13489718898/job/37697863328)

### Feature scope:
 
- [x] updated available model list for orchestration and openAI
- [x] updated ScenarioTest to check for deprecation of models

## Definition of Done

- [ ] ~Documentation updated~
- [ ] ~Release notes updated~


CharlesDuboisSAP (commented):
need release notes, maybe add a `@deperecated This model is not usable anymore` tag in the Javadoc

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiModel.java
```diff
@@ -26,6 +26,7 @@ public record OpenAiModel(@Nonnull String name, @Nullable String version) implem
   public static final OpenAiModel GPT_35_TURBO_1025 = new OpenAiModel("gpt-35-turbo-0125", null);
 
   /** Azure OpenAI GPT-3.5 Turbo chat completions model */
+  @Deprecated
```
Jonas-Isr, line 29: I am not sure whether marking it as "deprecated" is the correct solution here, since the model seems to be functionally removed by OpenAI, right? I would rather delete the option completely and add a note to the release notes? (The breaking change is not our fault but we should make it clear that it does not work anymore.)

What do you think, @newtork?

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiModel.java
```diff
@@ -40,6 +41,12 @@ public record OpenAiModel(@Nonnull String name, @Nullable String version) implem
   /** Azure OpenAI GPT-4o Mini chat completions model */
   public static final OpenAiModel GPT_4O_MINI = new OpenAiModel("gpt-4o-mini", null);
 
+  /** Azure OpenAI o1 chat completions model */
+  public static final OpenAiModel O1 = new OpenAiModel("o1", null);
+
+  /** Azure OpenAI o3-mini chat completions model */
+  public static final OpenAiModel O3_MINI = new OpenAiModel("o3-mini", null);
```
Jonas-Isr, line 48: This is "just" a drive-by change, right? 

Personally, I do not like the super short names here (even thought its their official names). Maybe we can find better ones like `OPEN_AI_01`? But no super strong feelings.

CharlesDuboisSAP (commented):
Should we also deprecate models that are deprecated by AI Core but still usable?
https://me.sap.com/notes/3437766
@newtork

TillK17:
- replaced gpt-3.5-turbo-16k with gpt-4o in the affected orchestration tests

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/ScenarioTest.java
```diff
@@ -33,7 +33,8 @@ void openAiModelAvailability() {
     // get the models from the OpenAiModel class
     List<String> declaredOpenAiModelList = new ArrayList<>();
     for (Field field : declaredFields) {
-      if (field.getType().equals(OpenAiModel.class)) {
+      if (field.getType().equals(OpenAiModel.class)
+          && !field.isAnnotationPresent(Deprecated.class)) {
```
CharlesDuboisSAP, line 37: I suggest:
```java
  void openAiModelAvailability() {

    // Gather AI Core's list of available OpenAI models
    final var aiModelList = new ScenarioController().getModels().getResources();

    final var availableOpenAiModels =
        aiModelList.stream()
            .filter(model -> model.getExecutableId().equals("azure-openai"))
            .collect(
                () -> new HashMap<String, Boolean>(),
                (list, model) -> {
                  list.put(model.getModel(), isDeprecated(model));
                },
                HashMap::putAll);

    // Gather our declared OpenAI models
    Field[] declaredFields = OpenAiModel.class.getFields();

    // get the models from the OpenAiModel class
    HashMap<String, Boolean> declaredOpenAiModelList = new HashMap<>();
    for (Field field : declaredFields) {
      if (field.getType().equals(OpenAiModel.class)) {
        declaredOpenAiModelList.put(
            ((OpenAiModel) field.get(null)).name(), field.isAnnotationPresent(Deprecated.class));
      }
    }

    // Assert that the declared OpenAI models match the expected list
    assertThat(declaredOpenAiModelList.keySet()).containsAll(availableOpenAiModels.keySet());
    for (var model : availableOpenAiModels.entrySet()) {
      Boolean declaredDeprecated = declaredOpenAiModelList.get(model.getKey());
      assertThat(declaredDeprecated)
          .withFailMessage(
              "%s is deprecated:%s on AI Core but deprecated:%s in AI SDK",
              model.getKey(), model.getValue(), declaredDeprecated)
          .isEqualTo(model.getValue());
    }
  }

  private static boolean isDeprecated(AiModelBaseData model) {
    Optional<AiModelVersion> version =
        model.getVersions().stream().filter(AiModelVersion::isIsLatest).findFirst();
    if (version.isEmpty()) {
      throw new RuntimeException();
    }
    return version.get().isDeprecated();
  }
```

jjtang1985 (dismissed):

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/ScenarioTest.java
newtork: > openAiModelAvailability()

Please note: we may not have all models in our available scenarios

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiModel.java
Jonas-Isr: After discussion I will mark this as deprecated to avoid breaking changes.

newtork:
As discussed we'll propagate deprecation notices, may attach date hints.
But we'll not remove enum values.

If our current embedding model is considered deprecated and we cannot simply switch, please consider creating a BLI to track progress and alignment with JS.

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/ScenarioTest.java
```diff
@@ -3,18 +3,21 @@
...
 import lombok.SneakyThrows;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
 
 class ScenarioTest {
 
   @Test
   @DisplayName("Declared OpenAI models must match AI Core's available OpenAI models")
+  @DisabledIfSystemProperty(named = "aicore.landscape", matches = "canary")
```
Jonas-Isr, line 20: This is to avoid running this test on canary, where more models might (and currently are) deployed than on prod.

CharlesDuboisSAP: Unfortunately [prod is yet not enabled](https://github.com/SAP/ai-sdk-java/blob/9cc853f49cf9c39d64da21c70a01fa21fb8c2a0a/.github/workflows/e2e-test.yaml#L17), this will disable the test

CharlesDuboisSAP (approved):

newtork (approved):
After discussing with PO we plan to migrate this test from release-blocking to scheduled workflow action, ideally to create PRs automatically for updating model list and deprecation state.

PR #350, opened 2025-02-26
chore: Enable E2E test for OpenAI models

newtork:
## Context

Related https://github.com/SAP/ai-sdk-java/pull/349


Jonas-Isr (approved):
LGTM

Jonas-Isr:
Should we add deprecation notices to `gpt-35-turbo-16k` and `text-embedding-ada-002`?

newtork:
> Should we add deprecation notices to `gpt-35-turbo-16k` and `text-embedding-ada-002`?

Let's look into this with #348

Issue #241, opened 2024-12-17
Is RAG (document grounding) feature integrated in Spring AI?


zxcvboy:
### Ask the Question

Dear colleague. 

    Is RAG feature supported by ai-sdk-java? If not, i there any roadmap? 

    Thanks a lot.

jjtang1985:
Hi @zxcvboy ,

Could you please share more details about your use cases?
- using the SharePoint with [pipeline API](https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/pipeline-api)
- adding embeddings with [vector API](https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/vector-api)

We planned to support the grounding [features](https://help.sap.com/docs/sap-ai-core/sap-ai-core-service-guide/grounding-3dd1bc414c1848fe869e94902a3ead9f) early next year.

zxcvboy:
Hi. @jjtang1985 

  I want to retrieve some results from my vector engine and send those results with my prompt to LLM.   
  Below is my sample code with Spring AI.  I just wonder if this feature can be achieved by ai-sdk-java or not.
------------ Spring AI sample -------------------------

  private final VectorStore vectorStore; // Spring AI自動配置
  private final ChatClient chatClient; // 使用Config類別建立好Bean,這裡可以直接綁定

 public String search2(String query) throws IOException {

		String result = chatClient.prompt().advisors(
				new QuestionAnswerAdvisor(vectorStore) // RAG增強器
		).user(query).call().content();

		return result;
	}

MatKuhr:
Yes, you can achieve the same RAG flow with the AI SDK. However, currently you have to implement the embedding calls and vector database retrieval manually. We have it on the roadmap to provide an integration for Spring AI, but this isn't completed yet.

You can refer to the [sample code](https://github.com/SAP/ai-sdk-java/blob/main/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OpenAiController.java#L195) on how to compute an embedding.

zxcvboy:
Hi. @MatKuhr.

   Appreciate for your reply. Is there a specific roadmap for integration with Spring AI? RAG is the most popular AI scenarios for SAP customers. Is it possible to provide a sample or scenario for a RAG use case? 

Thanks a lot.

MatKuhr:
[...]

PR #352, opened 2025-02-26
chore: Rename `grounding` to `document-grounding`

newtork:
## Context

PO made us aware we should change the new module name, before release.

I kept the java class namespace `com.sap.ai.sdk.grounding` as long as no-one complains.

- [x] Add missing release note


newtork:
https://github.com/SAP/ai-sdk-java/issues/241

PR #381, opened 2025-03-11
chore: Improve Generator Script

MatKuhr:
## Context

Minor improvement for our generator

### DIFF
--- a/.github/workflows/spec-update.yaml
+++ b/.github/workflows/spec-update.yaml
@@ -52,6 +52,18 @@ jobs:
         with:
           token: ${{ secrets.BOT_SDK_JS_FOR_DOCS_REPO_PR }}
 
+      - name: "Setup Java"
+        uses: actions/setup-java@v4
+        with:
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/sap/ai/sdk/core/model/KpiResultRowItem.java
```diff
@@ -11,123 +11,8 @@
...
-    return o.toString().replace("\n", "\n    ");
-  }
-
-  /** Create a new {@link KpiResultRowItem} instance. No arguments are required. */
-  public static KpiResultRowItem create() {
-    return new KpiResultRowItem();
-  }
-}
+@Beta
+public interface KpiResultRowItem {}
```
MatKuhr, line 18: Technically, this generated code isn't correct, but neither was the previous one, so not a breaking change.

According to the spec, this should implement `anyOf: [string, nunmber]`

newtork (approved):

PR #384, opened 2025-03-12
chore: Update available modellist

TillK17:
## Context

AI/ai-sdk-java-backlog#206

- added new models o1 and o3-mini to `OrchestrationAiModel` and `OpenAiModel`
- tested new parameter `reasoning effort` of o-series in bruno and SDK (works well, with high effort ~double the token consumption of low effort)
- deprecated models according to SAP Note and added retirement date and suggested replacement model to javadoc

### Feature scope:
 
- [x] All available models from the SDK are aligned with the SAP notes.
- [x] Both OpenAI client and orchestration client are covered
- [x] Test if new, additional model params for the new models already work. If not, create a follow-up ticket

## Definition of Done

- [x] Documentation updated
- [ ] ~Release notes updated~


foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiModel.java
```diff
@@ -16,53 +16,57 @@
...
-   * @deprecated This model is not usable anymore. It is retired on AI Core since 2025-02-13.
+   * @deprecated This model is deprecated on AI Core with a planned retirement on 2025-02-13. The
+   *     suggested replacement model is {@link OpenAiModel#GPT_4O_MINI}.
    */
   @Deprecated
   public static final OpenAiModel GPT_35_TURBO_16K = new OpenAiModel("gpt-35-turbo-16k", null);
 
-  /** Azure OpenAI GPT-4 chat completions model */
+  /** Azure OpenAI GPT-4 model */
   public static final OpenAiModel GPT_4 = new OpenAiModel("gpt-4", null);
```
rpanackal, line 49: (Question)

Isn't `GPT_4` also deprecated?

TillK17: `GPT_4.withVersion("0613")` is deprecated but `GPT_4.withVersion("turbo-2024-04-09  (latest)")` isnt. I wasnt quite sure how to handle this and believe there might be a mistake by Orchestration. They should've release `gpt-4-turbo` as an individual model instead of as a version of `gpt-4`. But with this version not being deprecated, I decided to not deprecate `GPT_4`

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModel.java
```diff
@@ -99,26 +105,45 @@ public class OrchestrationAiModel {
...
-  /** Azure OpenAI GPT-4 chat completions model */
+  /** Azure OpenAI GPT-4 model */
   public static final OrchestrationAiModel GPT_4 = new OrchestrationAiModel("gpt-4");
 
-  /** Azure OpenAI GPT-4-32k chat completions model */
+  /**
+   * Azure OpenAI GPT-4 model
+   *
+   * @deprecated This model is deprecated on AI Core with a planned retirement on 2025-05-13. The
+   *     suggested replacement model is {@link OrchestrationAiModel#GPT_4O}.
```
newtork, line 144: (Question)

Just to confirm, we want to expose this dynamic information (deprecation date and replacement), rather than suggesting SAP Note directly? @jjtang1985?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModel.java
```diff
@@ -127,6 +152,12 @@ public class OrchestrationAiModel {
   /** Azure OpenAI GPT-4o-mini chat completions model */
   public static final OrchestrationAiModel GPT_4O_MINI = new OrchestrationAiModel("gpt-4o-mini");
 
+  /** Azure OpenAI o1 chat completions model */
```
newtork, line 155: (Question)

Why do you remove chat completion everywhere but add it here?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModel.java
jjtang1985: This was initially proposed by Till.  
From consumer perspective, it's convenient to have this information as Java annotation.
If it does not cost too much time, I would like to have them.

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiModel.java
rpanackal: (Comment)

Seems like `GPT_4` _latest_ is also soon to be deprecated/even retired.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationAiModel.java
TillK17: True, I believe I wasnt 100% consistent with this. 

I initially noticed that some models had a _chat completion_ or _embedding_ in their JavaDoc, others not. As many of the newer models are generally multimodal, I found the _chat completion_ a bit misleading. However, for our SDK, they can of course only generate text or cumpute embeddings (except from dall-e).

rpanackal (approved):
LGTM

PR #391, opened 2025-03-19
fix: [Orchestration] default values for template messages

### LINKED ISSUE #389: How to use TemplateConfig...withDefaults?
### Ask the Question

I am creating a template for Chat completion of 2 messages
- system
```markdown
# Instruction

{{?system_instruction_text}}
...
```
- user
```markdown
# Input data
## Param1
`{{?param1}}
## Param2
`{{?param2}}
```

The user can override system_instruction_text parameter.
If I provide the param value via
```java
return TemplateConfig.create()
            .withTemplate(List.of(systemTemplate.createChatMessage(), userTemplate.createChatMessage()))
            .withDefaults(
                Map.of("system_instruction_text", systemPromptInstructionText)
            );
```
I receive error - unused parameter
If I provide the value at the very end with all other parameters, as in the [documentation](https://sap.github.io/ai-sdk/docs/java/guides/orchestration-chat-completion#chat-completion-with-templates), everything works.

My question: are these defaults for input parameters or something else?

Thank you.

newtork:
fixes #389

### DIFF
--- a/orchestration/src/main/java/com/sap/ai/sdk/orchestration/ConfigToRequestTransformer.java
+++ b/orchestration/src/main/java/com/sap/ai/sdk/orchestration/ConfigToRequestTransformer.java
@@ -51,20 +51,30 @@ final class ConfigToRequestTransformer {
     if (config instanceof TemplateRef) {
       return config;
     }
+
     val template = config instanceof Template t ? t : Template.create().template();
     val messages = template.getTemplate();
     val responseFormat = template.getResponseFormat();
     val messagesWithPrompt = new ArrayList<>(messages);
+
     messagesWithPrompt.addAll(
         prompt.getMessages().stream().map(Message::createChatMessage).toList());
     if (messagesWithPrompt.isEmpty()) {
       throw new IllegalStateException(
           "A prompt is required. Pass at least one message or configure a template with messages or a template reference.");
     }
-    return Template.create()
-        .template(messagesWithPrompt)
-        .tools(template.getTools())
-        .responseFormat(responseFormat);
+
+    val result =
+        Template.create()
+            .template(messagesWithPrompt)
+            .tools(template.getTools())
+            .defaults(template.getDefaults())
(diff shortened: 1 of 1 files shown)


orchestration/src/main/java/com/sap/ai/sdk/orchestration/ConfigToRequestTransformer.java
```diff
@@ -51,20 +51,30 @@ static TemplatingModuleConfig toTemplateModuleConfig(
...
-        .template(messagesWithPrompt)
-        .tools(template.getTools())
-        .responseFormat(responseFormat);
+
+    val result =
+        Template.create()
+            .template(messagesWithPrompt)
+            .tools(template.getTools())
+            .defaults(template.getDefaults())
+            .responseFormat(responseFormat);
```
rpanackal, line 72: (Question)

Is it viable to reuse the incoming template object instead of recreating one. 

We are now repeating the conversion to low level in `OrchestrationTemplate`.

newtork: i don't get the question, do you want me to add a unit test to improve coverage?

jjtang1985: As a fix, I would have a unit test for covering the case.

rpanackal: Yes a unit test works.

I was just thinking of how to avoid creating the low level object of `Template` **multiple times** like

```java
Template.create()
        .template(template)
        .defaults(defaults) // we forgot this in one location
        .responseFormat(responseFormat)
        .tools(tools);
```
so that we don't forget to initialise the object correctly in one of the time. I don't have a good solution to propose.

rpanackal (approved):
LGTM

PR #395, opened 2025-03-26
fix: HTML homepage collapse

CharlesDuboisSAP:
## Context

The orchestration container was too big, now every item is collapsed but the first one.


rpanackal (approved):
LGTM

PR #403, opened 2025-04-03
test: SpringAI Filtering 

TillK17:
## Context

AI/ai-sdk-java-backlog#224

- Input and output filtering for orchestration using the SpringAI Integration is tested as part of the sample app
- Example added to the documentation portal with [this PR](https://github.com/SAP/ai-sdk/pull/32)

### Feature scope:
 
- [x] content filter modules of the orchestration service supported in our Spring AI client
- [x] Sample code for content filtering
  - [x] Input filtering hits
  - [x] Output filtering hits
  - [ ] (optional) Streaming enabled

## Definition of Done

- [ ] ~Release notes updated~


sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/SpringAiOrchestrationTest.java
```diff
@@ -58,6 +58,22 @@ void testMasking() {
     assertThat(response.getResult().getOutput().getText()).isNotEmpty();
   }
```
TillK17, line 60: Tried to write minimal tests resistant to potential future changes in error messages. Not sure if this is the best way though

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/SpringAiOrchestrationService.java
```diff
@@ -94,6 +99,62 @@ public ChatResponse masking() {
...
+  @Nonnull
+  public ChatResponse inputFiltering(AzureFilterThreshold policy)
+      throws OrchestrationClientException {
+    val filterConfig =
+        new AzureContentFilter().hate(policy).selfHarm(policy).sexual(policy).violence(policy);
+    val opts =
+        new OrchestrationChatOptions(
+            config
+                .withLlmConfig(GEMINI_1_5_FLASH.withParam(TEMPERATURE, 0.0))
+                .withInputFiltering(filterConfig));
```
TillK17, line 120: used Gemini as the internal filter of OpenAI models is quite problematic for orchestration filter testing. To align with the normal orchestration service, we could consider using gemini as the 'standard' model for most of the endpoints?

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/SpringAiOrchestrationTest.java
```diff
@@ -58,6 +60,62 @@ void testMasking() {
...
+
+    var filterResult =
+        ((OrchestrationSpringChatResponse) response)
+            .getOrchestrationResponse()
+            .getOriginalResponse()
+            .getModuleResults()
+            .getOutputFiltering();
+    assertThat(filterResult.getMessage()).containsPattern("0 of \\d+ choices failed");
+  }
+
```
TillK17, line 118: Strict test for hitting filters and lenient tests for non-hitting filters

orchestration/src/main/java/com/sap/ai/sdk/orchestration/spring/OrchestrationSpringChatResponse.java
```diff
@@ -48,7 +49,12 @@ static List<Generation> toGenerations(@Nonnull final LLMModuleResultSynchronous
   }
 
   @Nonnull
-  static Generation toGeneration(@Nonnull final LLMChoice choice) {
+  static Generation toGeneration(@Nonnull final LLMChoice choice)
+      throws OrchestrationClientException {
+    if ("content_filter".equals(choice.getFinishReason())) {
+      throw new OrchestrationClientException("Content filter filtered the output.");
+    }
+
```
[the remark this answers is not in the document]
TillK17: More context: 
image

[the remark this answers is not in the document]
TillK17: The current integration of SpringAI doesnt allow for this, as the SpringAI `ChatResponse` is rebuild from the generated model classes of Orchestration. From my understanding this is necessary to get all properties of the response, as the convenience API abstracts away some details, but not completely sure. Maybe @CharlesDuboisSAP can give a more qualified answer.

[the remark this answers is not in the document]
TillK17: So I guess you could retrieve the response directly in the constructor and pass it separately to the toGenerations() or pass the `orchestrationResponse` to toGeneration() extracting the low-level `originalResponse()` there but I'm not sure if this is really better.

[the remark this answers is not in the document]
CharlesDuboisSAP: I think Spring AI doesn't throw on content filter. neither on `call()` or on `getText()`. You shouldn't expect any Exception.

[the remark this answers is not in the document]
TillK17: Hmm ok. I honestly have kind of a different understanding what a content filter should do, but this is then also rather a problem in orchestration IMO. If we say we only want to reflect the state of SpringAI then of course throwing an exception is wrong. But this is then basically also the case for "normal" orchestration, right? If orchestration says yeah the output filter was activated but I still give you the sensitive content of the output, then we should probably also not undermine this behaviour by throwing an exception in `getContent()`, to be consistent. 

But for me this is then not a content filter but rather giving little disclaimer in some JSON field no one ever looks at...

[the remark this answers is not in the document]
CharlesDuboisSAP: I read the docs + searched their code repo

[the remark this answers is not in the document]
TillK17: > Plus no matter what we don't allow for behaviour changes in released API.

I thought this would be ok as SpringAI is marked as Beta. 

Anyways, I can adapt the tests and only check for the finishReason. However, I'm personally not super convinced by this approach. It maybe depends on whether we see the SpringAI integration as a convenience or not. IMO, if we see it as a convenience layer, we should add the convenience to also throw an exception, like we did for normal orchestration. And then its maybe also not that important to stick to how SpringAI does things. If its no convenience, then I'm fine with just replicating the state of SpringAI.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/SpringAiOrchestrationController.java
```diff
@@ -52,6 +59,52 @@ Object template(@Nullable @RequestParam(value = "format", required = false) fina
...
+      return ResponseEntity.internalServerError().body(msg);
+    }
+
+    if ("json".equals(format)) {
+      return ((OrchestrationSpringChatResponse) response)
+          .getOrchestrationResponse()
+          .getOriginalResponse();
+    }
+    return response.getResult().getOutput().getText();
+  }
```
rpanackal, line 106: As demonstrated in the test, the Orchestration API returns an empty string when output content filtering is triggered. And, the `OrchestrationClientException` is only thrown when `OrchestrationChatResponse().getContent()` is explicitly invoked.

Currently, the controller method assumes an exception will be raised on content filtering. However, this is false, because spring integration access content via a lower-level API - effectively bypassing the exception `OrchestrationClientException`.

---
(Questions)

@CharlesDuboisSAP

Should we keep the exception handling demo  and align the Spring integration to this behavior?

Or is this silent handling intentional in the integration, making exception handling unnecessary?

MatKuhr: for context, see the [discussion here](https://github.com/SAP/ai-sdk-java/pull/403#discussion_r2031280040) 😉 

Given that discussion, this should probably be changed to:

```java
response = service.outputFiltering(policy);
if (response.hasFinishReasons(Set.of("content_filter"))) {
    throw new OrchestrationClientException("Content filter");
}
```

.. or something similar

TillK17: Yeah, you're right. I just forgot to remove the exception handling in the sample app after removing it in the SpringAI integration. Is fixed with the latest commit.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/SpringAiOrchestrationService.java
TillK17: Maybe I didnt express myself clear (or forgot a comma between "Gemini, as"). 

When using OpenAI models with more explicit content for triggering the output filter, the internal Azure Content Filter hits and produces a 400 so we cant test the output filter anymore (see screenshot). From my observation, it depends on the request, so you probably can always find prompts where this is not the case.  However, Gemini has no internal content filter so theres no risk of accidentally triggering it. Thats why I would always prefer to not test Filtering with OpenAI models. 



Screenshot 2025-04-09 094028

Full exception message: Request failed with status 400 Bad Request and error message: 'The response was filtered due to the prompt triggering Azure OpenAI's content management policy. Please modify your prompt and retry. To learn more about our content filtering policies please read our documentation: https://go.microsoft.com/fwlink/?linkid=2198766

MatKuhr: Ah okay, yes that is the reason we should use models without an internal filter for this test 👍🏻 

While it's possible to set the orchestration content filter to be more strict than the internal azure filter, engineering a prompt that reliably passes the internal filter but fails the orchestration filter is unnecessarily tedious and unreliable.

rpanackal (dismissed):
LGTM. If you switch back to OpenAI model use, let me know so I can simply reapprove.

newtork (approved):

newtork:
I ran and tested it locally. Success 👍

PR #406, opened 2025-04-03
enable Reproducible Builds

hboutemy:
## Context

Activate Reproducible Builds: https://maven.apache.org/guides/mini/guide-reproducible-builds.html

### Feature scope:
 
- [x] add timestamp 
- [x] update jar plugin (do not rely on default version provided by Maven)

## Definition of Done

- [ ] Error handling created / updated & covered by the tests above
- [ ] Release notes updated

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,6 +57,7 @@
     <maven.compiler.release>17</maven.compiler.release>
     <maven.compiler.proc>full</maven.compiler.proc>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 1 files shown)


Jonas-Isr (approved):
For future reference: Since the actual release-commits are not present in the commit history (due to us using squash merges), in order to reproduce the release build, one has to use the commit that includes the release-commit and then revert some of the changes (should be enough to set the sdk version back to `x.y.z`, i.e., removing `-SNAPSHOT`).

To build the release, we use `mvn clean install -DskipTests`.

PR #420, opened 2025-04-16
chore: Fix the prompt registry history test

Jonas-Isr:
## Context

The fix in https://github.com/SAP/ai-sdk-java/pull/411 was not correct. When trying to avoid flakyness in the history test, I changed it from `= 2` to `>2` in that PR, but it should be `>=2`.


sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/PromptRegistryTest.java
```diff
@@ -78,7 +78,7 @@ void history() {
     PromptTemplateListResponse history = controller.history();
     // Bug that doesn't delete prompts fast enough. Should be equal to 2
     assertThat(history.getCount()).isGreaterThanOrEqualTo(2);
-    assertThat(history.getResources()).hasSizeGreaterThan(2);
+    assertThat(history.getResources()).hasSizeGreaterThanOrEqualTo(2);
```
Jonas-Isr, line 81: The issue is that the deletion in prompt registry takes very long (even though a positive response is sent immediately). Thus, often there is still history left here, so we only check `>=2` instead of `=2` here.

PR #423, opened 2025-04-24
feat: [Orchestration] Enable local prompt templates

Jonas-Isr:
## Context

AI/ai-sdk-java-backlog#215

This PR enables users to validate and use prompt templates from their local environment, without the use of prompt registry. For this, the user can provide a String containing a YAML representation of the prompt template and we will deserialise this into our (high level) `OrchestrationTemplate` class, which then can be used to configure an Orchestration call. If the provided prompt template does not match the `prompt-registry` spec, an exception is thrown. 

__Note 1:__ We have to assume that the `prompt-registry` spec keeps exactly equal to the `orchestration` spec for the relevant fields. If the two diverge, we can no longer deserialise the prompt template (which follows `prompt-registry` spec) into our Java classes (which follow mostly the `orchestration` prompt). Since we deserialise into high-level classes, we might have some wiggle room, but that would probably become quite messy.

__Note 2:__ The whole process on our side (i.e. validation and deserialisation) ignores the `additionalFields` of the prompt templates. This is aligned with PO and JS SDK. (The reason is that we cannot know what is inside these fields, and it usually is information we cannot handle in our template classes.)

### Feature scope:
 
- [x] Local yaml files can be used to create templates without using prompt registry
- [x] the yaml gets validated

## Definition of Done

  - PR: https://github.com/SAP/ai-sdk/pull/83
- [x] Release notes updated


orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationTemplate.java
```diff
@@ -93,4 +107,45 @@ public OrchestrationTemplate withJsonResponse() {
...
+      obj = yamlReader.readValue(inputYaml, Object.class);
+    } catch (JsonProcessingException ex) {
+      throw new IOException("Failed to parse the YAML input: " + ex.getMessage(), ex);
+    }
+    try {
+      final ObjectMapper jsonWriter = new ObjectMapper();
+      return fromJson(jsonWriter.writeValueAsString(obj));
+    } catch (JsonProcessingException ex) {
+      throw new IOException("Failed to deserialize the input: " + ex.getMessage(), ex);
+    }
```
Jonas-Isr, line 149: I am not sure if these are the correct/best Exception types. I am open to suggestions :)

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationTemplate.java
```diff
@@ -93,4 +107,45 @@ public OrchestrationTemplate withJsonResponse() {
...
+   * @return A Template object representing the provided JSON
+   * @since 1.7.0
+   */
+  @Nullable
+  private OrchestrationTemplate fromJson(@Nonnull final String inputString) throws IOException {
+    final ObjectMapper objectMapper =
+        OrchestrationJacksonConfiguration.getOrchestrationObjectMapper();
+    final JsonNode rootNode = objectMapper.readTree(inputString);
+    return objectMapper.treeToValue(rootNode.get("spec"), OrchestrationTemplate.class);
+  }
```
Jonas-Isr, line 125: This method could principally be made public as an additional feature. I left it private because it is not part of the ticket, not tested on its own, and not aligned with JS SDK (which only has the yaml feature afaik).

orchestration/pom.xml
```diff
@@ -108,6 +108,10 @@
       <groupId>com.github.victools</groupId>
       <artifactId>jsonschema-module-jackson</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.dataformat</groupId>
+      <artifactId>jackson-dataformat-yaml</artifactId>
+    </dependency>
```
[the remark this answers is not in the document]
Jonas-Isr: We need this to deserialise yaml files. (Note that it is dat**a**format, not dat**e**format.) 

Explicitly, we need `com.fasterxml.jackson.dataformat.yaml.YAMLFactory` to create an `ObjectMapper` in the `fromYaml()` method of `OrchestrationTemplate`.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationTemplate.java
```diff
@@ -93,4 +107,45 @@ public OrchestrationTemplate withJsonResponse() {
...
+  /**
+   * Create a {@link Template} object from a YAML provided as String.
+   *
+   * @throws IOException if the YAML cannot be parsed or deserialized
+   * @param inputYaml the provided YAML
+   * @return A Template object representing the provided YAML
+   * @since 1.7.0
+   */
+  @Nullable
+  public OrchestrationTemplate fromYaml(@Nonnull final String inputYaml) throws IOException {
```
Jonas-Isr, line 136: (Question to reviewer)
Should we maybe annotate this with `@Beta`? 

Pro: We might not be able to continue to keep this method/feature alive without additional work if the specs of prompt-registry and orchestration were to diverge at some point (see my comment in the PR description).

Con: Additional beta annotation for convenience code.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationTemplate.java
newtork: Instead of using two objectmappers, I was about to pitch something like...

```java
  private static class TemplateYaml {
    @JsonProperty("spec")
    private OrchestrationTemplate spec;
  }
```

But that would've required additional customization to propagate the modules/rules along for Yaml Factory.

orchestration/src/test/java/com/sap/ai/sdk/orchestration/OrchestrationUnitTest.java
```diff
@@ -978,4 +980,46 @@ void testTemplateFromPromptRegistryByScenario() throws IOException {
...
   }
+
+  @Test
+  void testTemplateFromInput() throws IOException {
+    stubFor(
+        post(anyUrl())
+            .willReturn(
+                aResponse()
+                    .withBodyFile("templateReferenceResponse.json")
+                    .withHeader("Content-Type", "application/json")));
```
newtork, line 991: That's a little indentation gore :)

sample-code/spring-app/pom.xml
```diff
@@ -170,6 +170,10 @@
           <groupId>ch.qos.logback</groupId>
           <artifactId>logback-classic</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.yaml</groupId>
+          <artifactId>snakeyaml</artifactId>
+        </exclusion>
```
newtork, line 176: Exclusion is not the recommended way to fix dependency-convergence errors. You can find in SAP Cloud SDK, we use dependencyManagement to override versions for all transitive occurrences
https://github.com/SAP/cloud-sdk-java/blob/main/pom.xml

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
```diff
@@ -270,4 +274,18 @@ Object templateFromPromptRegistryByScenario(
...
     return response.getContent();
   }
+
+  @GetMapping("/localPromptTemplate")
+  @Nonnull
+  Object localPromptTemplate(@RequestParam(value = "format", required = false) final String format)
+      throws IOException {
+    final var resource = resourceLoader.getResource("classpath:promptTemplateExample.yaml");
+    final var promptTemplate =
+        new String(resource.getInputStream().readAllBytes(), StandardCharsets.UTF_8);
```
newtork, line 284: (Minor)

This looks a little ugly, have you checked the `@Value` annotation and resource handling?

```java
    @Value("classpath:myfile.txt")
    private Resource resource;

    @GetMapping("/localPromptTemplate")
    Object foo() {
        String content = Files.readString(resource.getFile().toPath(), UTF_8);
```

orchestration/src/test/java/com/sap/ai/sdk/orchestration/OrchestrationUnitTest.java
Jonas-Isr: That is true 😅 But the code-style plugin seems to really like it that way...

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationTemplate.java
newtork: as discussed, since this API has no alternative to users, I would keep it non-Beta

newtork (approved):

PR #460, opened 2025-06-04
fix: [Orchestration] spec update

CharlesDuboisSAP:
## Context

AI/ai-sdk-java-backlog#448.

Update to the latest spec from orchestration [from this branch](https://github.tools.sap/AI/llm-orchestration/pull/1463)

### Feature scope:
 
- [x] Split the response into `synchronous` and `streaming` instead of a `oneof`
- [x] DPI
- [x] Azure
- [x] Embedding 
- [x] Prompt Shield
- [x] Remove "Synchronous" suffix

## Definition of Done

- [x] [Documentation](https://github.com/SAP/ai-sdk/tree/main/docs-java) updated -> https://github.com/SAP/ai-sdk/pull/156
- [x] Release notes updated

### DIFF
--- a/.github/workflows/spec-update.yaml
+++ b/.github/workflows/spec-update.yaml
@@ -88,8 +88,8 @@ jobs:
               FILE_PATH='core-services/document-grounding/src/main/resources/spec/grounding.yaml'
               ;;
             orchestration)
-              API_URL="$API_BASE_URL/AI/llm-orchestration/contents/src/spec/api.yaml?ref=$REF"
-              FILE_PATH='orchestration/src/main/resources/spec/v1.yaml'
+              API_URL="$API_BASE_URL/AI/llm-orchestration/contents/src/spec/v2.yaml?ref=$REF"
+              FILE_PATH='orchestration/src/main/resources/spec/orchestration.yaml'
               ;;
             prompt-registry)
               API_URL="$API_BASE_URL/AI/prompt-registry/contents/src/spec/prompt-registry.yaml?ref=$REF"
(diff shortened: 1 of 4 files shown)


orchestration/src/main/java/com/sap/ai/sdk/orchestration/ContentFilter.java
```diff
@@ -17,11 +18,20 @@
...
-   * A method that produces the serializable equivalent {@link FilterConfig} object from data
+   * A method that produces the serializable equivalent {@link InputFilterConfig} object from data
    * encapsulated in the {@link ContentFilter} object.
    *
-   * @return the corresponding {@code FilterConfig} object.
+   * @return the corresponding {@link InputFilterConfig} object.
    */
   @Nonnull
-  FilterConfig createConfig();
+  InputFilterConfig createInputFilterConfig();
```
newtork, line 27: (Comment)

While I agree to the solution, it doesn't feel great that we're changing public API on convenience layer. That would be considered a breaking change normally. I do understand we only use it as workaround to expose the low-level converter method(s), but still.. 🤔

CharlesDuboisSAP: I would have blocked the generation entirely until they reverted the spec, but I was on vacation and you accepted the breaking change. I'm still in favor of blocking changes that don't compile.

newtork (commented):
Looks good. Do not merge yet.
As of `2025-07-02` version [0.77.3](https://github.tools.sap/AI/llm-orchestration/blob/v0.77.3/src/spec/api.yaml) is active in production.

<details><summary>There's still a spec diff, we have to wait out</summary>

image


</details>

<details><summary>0.81.4 is in pre-production with almost no changes necessary</summary>

image

</details>

PR #495, opened 2025-07-11
chore: [DevOps] fix e2e test

CharlesDuboisSAP:
We added `promptShield` which means filters are not skipped anymore


Jonas-Isr (approved):
🛡️

PR #543, opened 2025-08-15
chore: [Orchestration] reduce exception types

newtork:
## Context

Related
* Alternative https://github.com/SAP/ai-sdk-java/pull/537
* Discussion https://sap-ml.slack.com/archives/C070WHEH35L/p1755089417470779

Instead of making the API more strict by adding exception types for different error cases, we can also reduce it to just one.

### Feature scope:
 
- [x] Similar to OpenAI exception type(s)
- [x] Exception refactoring
  ```diff
    ClientException
      OrchestrationClientException
  -     OrchestrationFilterException
  -       Input
  -       Output
  ```
- [x] No breaking changes, we haven't released the experimental exception classes yet.

## Definition of Done

- [x] ~[Documentation](https://github.com/SAP/ai-sdk/tree/main/docs-java) updated~
- [x] ~Release notes updated~


rpanackal (commented):
(Major)

I don't see how the user can tell whether its the input or output filter that failed.

```java
try {
   response = client.chatCompletion(prompt, configWithIOFilter)
   content = response.getContent();
} catch (OrchestrationClientException e) {
 // ...
}
```

But I would expect something like `isInputFilterException()` or  the like.. so the user would know which methods to call afterwards.

Currently, the user would have to `try-catch` the call to `chatCompletion()` and `getContent()` seperately.

rpanackal (commented):
Overall, I agree with Charles that this approach is preferred over the one in [chore: Improve exception accessors](https://github.com/SAP/ai-sdk-java/pull/537).

PR #618, opened 2025-10-15
chore: [DevOps] Rename `bug report` to `problem report`

CharlesDuboisSAP:
Multiple customer are confused about reporting a problem, they should use the `bug report` issue template.
I renamed it `problem report` to make it less confusing.


CharlesDuboisSAP:
> I also think that removing the `question` template is fine :)

Yeah, you can create a blank issue anyway

PR #794, opened 2026-03-25
feat: [OpenAI] PoC - Responses API support with OpenAI SDK Adapter

rpanackal:
## Context

AI/ai-sdk-java-backlog#364.

This PoC provides an adapter for Official OpenAI SDK integrations with our SDK by implementing `com.openai.core.http.HttpClient`. You can find out more about the OpenAI recommended approach [here in their docs](https://github.com/openai/openai-java?tab=readme-ov-file#custom-http-client).

### Feature scope:
 
- [x] Introduces a new generic `AiCoreOpenAiClient` that can work as an adapter for any OpenAI endpoints
       - Easy adoption of any openai endpoints (eg: `/realtime`) that are/will be supported by AiCore 
- [x] Introduce `AiCoreHttpClientImpl`
  - [x] Implements a new request and response handling pathway (incl. streaming) that is compatible with OpenAI SDK.
  - [x] Returns a client with familiar OpenAI SDK API usage for the user.
- [x] Constrains allowed endpoint to `/response` (for now)

### Usage

```java
import com.openai.client.OpenAIClient
import com.sap.ai.sdk.foundationmodels.openai.v1.AiCoreOpenAiClient

OpenAiClient client = AiCoreOpenAiClient.forModel(GPT_5)
var request = ResponseCreateParams.builder().input(input).model(ChatModel.GPT_5).build()
Response response = client.responses().create(request)
```

#### Pros
-  Usage via official OpenAI SDK's API; minimal learning curve
-  Minimal maintenance overhead as we dont need to maintain models as well as client api.
- Customisability and ease of integration: By relaxing allowed method, endpoint etc, we can easy support additional openai endpoints (eg `/chat/completion`, `/realtime` etc)

#### Cons
- More complex implementation that introduces additional request/response handling pathway in addition to the existing ones
- Two different classes for AI Model list which can be confusing
  -  `com.sap.ai.sdk.foundationmodels.openai.OpenAiModel` used to select available models in AICore and `com.openai.models.ChatModel.ChatModel` for request payload configuration in OpenAI SDK
  - We can't pick one without loss of semantics

## Definition of Done

- [x] Functionality scope stated & covered
- [x] Tests cover the scope above
- [x] Error handling created / updated & covered by the tests above
- [x] Aligned changes with the JavaScript SDK
- [ ] [Documentation](https://github.com/SAP/ai-sdk/tree/main/docs-java) updated
- [x] Release notes updated

PR #806, opened 2026-03-30
feat: [OpenAI] PoC - An AiCore wrapper for OpenAi implementation

rpanackal:
## Context

AI/ai-sdk-java-backlog#364.

This is in reference to the [comment](https://github.com/SAP/ai-sdk-java/pull/794#discussion_r3000303132) in related PR.

Builds on the HTTP adapter approach and adds a service-level interception layer to handle the `"model"` field transparently.

**Problem:** the OpenAI SDK's request params require a `model` field in the body — meaning users are forced to specify the model twice, once when resolving the deployment and once in the request parameters. Omitting it causes a server-side error. 

The problem can be generalized to bring up the fact that we have limited control on OpenAI SDK usage as of before.

**Solution:** SDK service-level delegation
`AiCoreOpenAiClient.forModel(...)` returns a standard `com.openai.client.OpenAIClient`. Internally, its service accessors (e.g. `responses()`) return AI Core specific wrapper implementations (`AiCoreResponsesService`) that intercept each `create()`/`createStreaming()` call __before serialization__ at the Java API layer.

The model resolved at client creation time is __injected automatically__ into the request params if absent. If the user explicitly sets a model that does not match the deployed model, a descriptive error is thrown.

## Definition of Done

- [ ] ~[Documentation](https://github.com/SAP/ai-sdk/tree/main/docs-java) updated~
- [ ] ~Release notes updated~


newtork (commented):
It's good idea!

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/AiCoreChatCompletionService.java
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  @Override
+  @Nonnull
+  public ChatCompletion create(
+      @Nonnull final ChatCompletionCreateParams params,
+      @Nonnull final RequestOptions requestOptions) {
+    throwOnModelMismatch(params.model().asString());
+    return delegate.create(params, requestOptions);
+  }
```
rpanackal: Limitation 2:

In `AiCoreResponseService`, we are able to plug in the `model` ourselves, removing the burden from mentioning it twice. Once for deployment resolution and once in `ChatCompletionCreateParams` (otherwise server returns 400). 

The same is not true for `AiCoreChatCompletionService`. `ChatCompletionCreateParams` throws right away when `model` is not declared -> we can not provide any convenience to user by plugging in model downstream.

rpanackal (commented):
I am halting development using the solution proposed in this PR since its doesn't scale well for the value it provides currently.

Wrapping OpenAI SDK api along with each Services eg: `ResponseService`, `ChatCompletionService`, `ChatService`, considering the size of their public API surface is not feasible.

PR #812, opened 2026-04-02
chore: Use GH app instead of PAT

marikaner:
## Context

Related to https://github.com/SAP/ai-sdk-js-backlog/issues/522.

Update the workflows where `BOT_SDK_JS_FOR_DOCS_REPO_PR` was used to use either `GITHUB_TOKEN` or a github app token, where needed.

Notes for reviewers:
- `GITHUB_TOKEN` cannot be used, if a follow up workflow shall be triggered. => https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow
- GH workflow permissions only apply to the `GITHUB_TOKEN`. When using a different token permissions are managed differently. => https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions

### DIFF
--- a/.github/workflows/dependabot-automerge.yaml
+++ b/.github/workflows/dependabot-automerge.yaml
@@ -1,4 +1,4 @@
-name: "Dependabot Auto-Merge"
+name: 'Dependabot Auto-Merge'
 
 on:
   workflow_dispatch:
@@ -14,14 +14,22 @@ env:
 jobs:
   review-prs:
     runs-on: ubuntu-latest
-    permissions:
-      pull-requests: write
-      contents: write
     steps:
+      - name: Generate GitHub App token
+        id: app-token
+        uses: actions/create-github-app-token@v3
+        with:
+          app-id: ${{ secrets.SAP_AI_SDK_BOT_CLIENT_ID }}
+          private-key: ${{ secrets.SAP_AI_SDK_BOT_PRIVATE_KEY }}
+          permission-contents: write
+          permission-pull-requests: write
+
       - name: Checkout
         uses: actions/checkout@v6
 
       - name: Approve and Merge PRs
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
         run: |
           PRS=$(gh pr list --app "dependabot" --state "open" --json number,title)
           PR_NUMBERS=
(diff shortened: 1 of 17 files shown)


.github/workflows/dependabot-automerge.yaml
```diff
@@ -14,29 +14,37 @@ env:
 jobs:
   review-prs:
     runs-on: ubuntu-latest
-    permissions:
-      pull-requests: write
-      contents: write
```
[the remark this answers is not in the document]
CharlesDuboisSAP: [Failed to create token for "ai-sdk-java" (attempt 1): A JSON web token could not be decoded](https://github.com/SAP/ai-sdk-java/actions/runs/24386755934/job/71222342624)

.github/workflows/spec-update-cleanup.yaml
```diff
@@ -34,7 +34,6 @@ on:
 jobs:
   cleanup:
     name: 'Delete Spec Update Branch'
-    needs: [auth]
```
CharlesDuboisSAP, line 37: Invalid workflow file: .github/workflows/spec-update-cleanup.yaml#L1
[(Line: 35, Col: 3): The workflow must contain at least one job with no dependencies.](https://github.com/SAP/ai-sdk-java/actions/runs/24386585964)

Also valid for `weekly-spec-update-cleanup.yaml`
Also this is missing a name.

.github/workflows/spec-update.yaml
```diff
@@ -62,10 +59,21 @@ jobs:
...
+      - name: 'Generate GitHub App token'
+        id: app-token
+        uses: actions/create-github-app-token@v3
+        with:
+          app-id: ${{ secrets.SAP_AI_SDK_BOT_CLIENT_ID }}
+          private-key: ${{ secrets.SAP_AI_SDK_BOT_PRIVATE_KEY }}
+          owner: SAP
+          repositories: ai-sdk-java
+          permission-contents: write
+          permission-pull-requests: write
```
CharlesDuboisSAP, line 71: [Failed to create token for "ai-sdk-java" (attempt 1): A JSON web token could not be decoded ](https://github.com/SAP/ai-sdk-java/actions/runs/24386988359/job/71223108639)

Also valid for every other workflow

.github/workflows/dependabot-automerge.yaml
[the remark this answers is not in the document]
marikaner: Ok, the error above is fixed now. Apparently I made a mistake when pasting the client ID.

.github/workflows/spec-update.yaml
marikaner: I made a mistake when creating the secrets. It is now fixed.

.github/workflows/spec-update-cleanup.yaml
marikaner: This should fix the problem, right?
What name do you mean?

CharlesDuboisSAP (approved):

PR #818, opened 2026-04-07
feat: [Orchestration] support the file uploading feature

n-o-u-r-h-a-n:
## Context

AI/ai-sdk-java-backlog#375.

Orchestration service releases a new service for file uploading(PDF), so the feature is now supported.

### Feature scope:
 
- [x] Supported feature by path locally.
- [x] Supported feature base64 encoded file as a string.
- [x] Added unit tests and end to end tests.
- [x] Coverage checks have slightly risen.

## Definition of Done

- [x] Release notes updated


sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/OrchestrationTest.java
```diff
@@ -376,6 +376,44 @@ void testMultiStringInput() {
...
   }
 
+  @Test
+  void testFileInputLocalPath() throws IOException {
+    final Path pdfPath = Files.createTempFile("orchestration-test", ".pdf");
+    Files.writeString(pdfPath, "%PDF-1.7\n%%EOF", StandardCharsets.UTF_8);
+
+    val result = service.fileInput(pdfPath).getOriginalResponse();
+    val choices = (result.getFinalResult()).getChoices();
+    assertThat(choices.get(0).getMessage().getContent()).isNotEmpty();
```
CharlesDuboisSAP:
```suggestion
    val response = service.fileInput(pdfPath);
    assertThat(response.getContent()).containsAnyOf("dummy", "Dummy");
```

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/OrchestrationTest.java
```diff
@@ -376,6 +376,44 @@ void testMultiStringInput() {
...
+    final HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+    connection.setRequestProperty("User-Agent", "Test implementation");
+
+    final String base64Data;
+    InputStream inputStream = connection.getInputStream();
+    base64Data = Base64.getEncoder().encodeToString(inputStream.readAllBytes());
+
+    val result = service.fileInputBase64(base64Data, "dummy.pdf").getOriginalResponse();
+    val choices = (result.getFinalResult()).getChoices();
+    assertThat(choices.get(0).getMessage().getContent()).isNotEmpty();
```
CharlesDuboisSAP:
```suggestion
    val response = service.fileInputBase64(base64Data, "dummy.pdf");
    assertThat(response.getContent()).containsAnyOf("dummy", "Dummy");
```

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/OrchestrationTest.java
```diff
@@ -376,6 +376,44 @@ void testMultiStringInput() {
...
+  }
+
+  @Test
+  void testFileInputUrl() {
+    final String fileUrl =
+        "https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf";
+
+    val result = service.fileInput(fileUrl, null).getOriginalResponse();
+    val choices = (result.getFinalResult()).getChoices();
+    assertThat(choices.get(0).getMessage().getContent()).isNotEmpty();
```
CharlesDuboisSAP:
```suggestion
    val response = service.fileInput(fileUrl, null);
    assertThat(response.getContent()).containsAnyOf("dummy", "Dummy");
```

CharlesDuboisSAP (commented):
The `index.html` and `OrchestrationController` are missing

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
```diff
@@ -85,6 +96,75 @@ public UserMessage withImage(@Nonnull final String imageUrl) {
...
+   * @since 1.18.0
+   */
+  @Nonnull
+  public UserMessage withFileBase64(
+      @Nonnull final String base64Data, @Nullable final String filename) {
+    validatePdfFilename(filename);
+    final String payload =
+        base64Data.startsWith(PDF_DATA_URI_PREFIX) ? base64Data : PDF_DATA_URI_PREFIX + base64Data;
+    try {
+      Base64.getDecoder().decode(payload.substring(PDF_DATA_URI_PREFIX.length()));
```
CharlesDuboisSAP:
```suggestion
```
Why did you decode it?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
```diff
@@ -85,6 +96,75 @@ public UserMessage withImage(@Nonnull final String imageUrl) {
...
+
+  @Nonnull
+  private UserMessage appendFileItem(
+      @Nonnull final String fileData, @Nullable final String filename) {
+    final var contentItems = new LinkedList<>(content.items());
+    contentItems.add(new FileItem(fileData, filename));
+    return new UserMessage(new MessageContent(contentItems));
+  }
+
+  private static void validatePdfFilename(@Nullable final String filename) {
```
CharlesDuboisSAP: Why did you create this function? Also it only gives a warn log even though it's called validate

orchestration/src/test/java/com/sap/ai/sdk/orchestration/OrchestrationConvenienceUnitTest.java
```diff
@@ -89,6 +95,99 @@ void testMessageConstructionImage() {
...
+            .role(USER);
+
+    assertThat(message.createChatMessage()).isEqualTo(expectedLowLevelMessage);
+  }
+
+  @Test
+  void testMessageConstructionFileFromNonPdf() {
+    assertThatThrownBy(() -> Message.user(Path.of("sample.txt")))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessageContaining("Failed to read the file: sample.txt");
```
CharlesDuboisSAP: This only fails because the file doesn't exists not because it's not a PDF

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
n-o-u-r-h-a-n: Alex has asked to make it only a warning instead of IllegalArgumentException as it might be better for the future when Orchestration support more file types.

I can rename the method perhaps?

CharlesDuboisSAP: Yes rename it

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
n-o-u-r-h-a-n: This line works on a substring from payload and decodes the raw base64 text to pdf bytes  , so that if it gets decoded successfully , the method will continue, if not it will go to the catch and throw an exception.

Also, the variable payload will not have its value changed and the new decoded bytes are not assigned to any variable, so no worries! 

What do you think about it?

orchestration/src/test/java/com/sap/ai/sdk/orchestration/OrchestrationConvenienceUnitTest.java
n-o-u-r-h-a-n: Okay I found a .txt file in the resources , so I used its path instead and removed the exception assertion as now the method which I renamed as Alex said only displays a warning not IllegalArgumentException.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/UserMessage.java
CharlesDuboisSAP: But why?

n-o-u-r-h-a-n: just to validate the base64 string , but could be removed if we assume that the base64 string entered by the user is always correct/valid

Jonas-Isr: While I think that validation is often a good idea, here it might not be necessary. For one, it will (hopefully) not often be relevant because users will use correct data. Also, we just talked with JS team and they also do not validate because the backend does it in a good way.

n-o-u-r-h-a-n: Okay, perfect then I'll just remove it now!

CharlesDuboisSAP (approved):

PR #822, opened 2026-04-10
chore: PoC wrapping official OpenAI SDK

ZhongpinWang:
## Context

AI/ai-sdk-java-backlog#ISSUENUMBER.

Try the sample app and call:

```
http://localhost:8080/openai-wrapper/chatCompletion/ai-sdk-java-e2e
```

The model call in OpenAI official client is not nice I agree, perhaps we can find a way in Java to hide it.

Ignore the `pom.xml` if the changes contain stupid stuff as they were AI generated but they work :)

### Feature scope:
 
- [ ] Task1 
- [ ] Task2 
  - [ ] SubTask1 

## Definition of Done

- [ ] Error handling created / updated & covered by the tests above
- [ ] Release notes updated

### DIFF
--- a/foundation-models/openai-wrapper/src/main/java/com/sap/ai/sdk/foundationmodels/openaiwrapper/OpenAiClientProvider.java
+++ b/foundation-models/openai-wrapper/src/main/java/com/sap/ai/sdk/foundationmodels/openaiwrapper/OpenAiClientProvider.java
@@ -1,0 +1,148 @@
+package com.sap.ai.sdk.foundationmodels.openaiwrapper;
+
+import com.openai.client.OpenAIClient;
+import com.openai.client.okhttp.OpenAIOkHttpClient;
+import com.openai.credential.BearerTokenCredential;
+import com.sap.ai.sdk.core.AiCoreService;
+import com.sap.ai.sdk.core.AiModel;
+import com.sap.cloud.sdk.cloudplatform.connectivity.DefaultHttpDestination;
+import com.sap.cloud.sdk.cloudplatform.connectivity.Header;
+import com.sap.cloud.sdk.cloudplatform.connectivity.HttpDestination;
+import java.util.Collection;
+import java.util.Locale;
+import javax.annotation.Nonnull;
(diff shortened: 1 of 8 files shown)


foundation-models/openai-wrapper/src/main/java/com/sap/ai/sdk/foundationmodels/openaiwrapper/OpenAiClientProvider.java
```diff
@@ -0,0 +1,148 @@
...
+    final DefaultHttpDestination defaultDest = (DefaultHttpDestination) destination;
+    final String baseUrl = defaultDest.getUri().toString() + "v1/";
+    log.debug("Building OpenAI client with base URL: {}", baseUrl);
+
+    return OpenAIOkHttpClient.builder()
+        .baseUrl(baseUrl)
+        .credential(BearerTokenCredential.create(() -> extractBearerToken(defaultDest)))
+        .putHeader("AI-Resource-Group", getResourceGroupHeader(defaultDest))
+        .build();
+  }
```
rpanackal, line 114: (Major)

Afaik, OpenAI SDK do not handle token refresh. So effectively, the obtained instance of `OpenAIOkHttpClient` also "expires" which was the concern I had when I started out here.

So, I add an adapter (custom `com.openai.core.http.HttpClient`) in my PR as we have a familiar Apache client that handles destination logic (incl token refresh)(https://github.com/SAP/ai-sdk-java/pull/794).

rpanackal (commented):
We are also good with the wrapper strategy on Java side. Even better since `model` and `api-version` will soon be unnecessary in v1. 

That said, currently, we still considering to release a Response API specific wrapper (checkout https://github.com/SAP/ai-sdk-java/pull/819). 

With a regular client instance, users can do the following

  ```java
  // supported
  client.chat().completion()
  client.response()
  
  //unsupported
  client.completion()
  client.realtime()
  client.audio()
  ```
  
  Ideally, we want to limit exposed api to what is supported.

PR #824, opened 2026-04-13
chore: Bump version of mpc-core to avoid CVE-2026-35568

Jonas-Isr:
## Context

This PR bumps the version of `io.modelcontextprotocol.sdk:mcp-core` to avoid [CVE-2026-35568](https://nvd.nist.gov/vuln/detail/CVE-2026-35568).


CharlesDuboisSAP (approved):
For now let's keep fixing and scanning the sample app.
If we at some point feel it gets overwhelming we can always remove it from the scan.

PR #832, opened 2026-04-21
fix: [Grounding] test

newtork:
`MatchMode` cannot be `null` ffs


Jonas-Isr:
> `MatchMode` cannot be `null` ffs

But it is `nullable: true` [here](https://github.com/SAP/ai-sdk-java/blob/bb1f9cb9353269dbda7077a1e4aa7e9beadd102d/core-services/document-grounding/src/main/resources/spec/grounding.yaml#L2336). Did you tell them that their spec is wrong? Or am I missing something?

Jonas-Isr (approved):

newtork:
Good catch! I updated my report in slack support channel

PR #833, opened 2026-04-21
fix: [PromptRegistry] `byte[]` to `File` on uploads

newtork:
- [x] Find proper solution

See diff:
- multipart payload requires `filename` definition

image

Currently the generated code will not offer that since internally it uses `ByteArrayBody` instead of `FileBody`.
The latter will add a `filename` reference.

image

Currently the workaround is to use `File` instead of `byte[]`.

---

## Update 7, May

Edit: This PR now holds the actual fix. Earlier, it served as an example for what the generated code should like.

Now, since the latest cloud sdk release (5.30.0), we have regenerated the prompt-registry module.


pom.xml
```diff
@@ -286,11 +286,19 @@
...
           <artifactId>openapi-generator-maven-plugin</artifactId>
           <version>${cloud-sdk.version}</version>
+          <dependencies>
+            <!-- Handle regression in 7.22.0 -->
+            <dependency>
+              <groupId>org.openapitools</groupId>
+              <artifactId>openapi-generator</artifactId>
+              <version>7.21.0</version>
+            </dependency>
+          </dependencies>
```
rpanackal, line 296: 7.22.0 brings in a regression on generated mode classes.

Image

Image

`String` (before) is now generated as `TrckExecutionId`. `TrckExecutionId` has a schema in spec and is simply a string container. Previously, such string container schemas were resolved to inner String for simplicity.

The resolution is breaking at the moment and there is no equivalent model class generated.

rpanackal: Related BLI: https://github.com/SAP/ai-sdk-java-backlog/issues/396

Jonas-Isr (approved):
👍

PR #834, opened 2026-04-23
chore: Add permissions

marikaner:
Should fix the CodeQL findings.

### DIFF
--- a/.github/workflows/perform-release.yaml
+++ b/.github/workflows/perform-release.yaml
@@ -20,6 +20,12 @@ env:
 jobs:
   prerequisites:
     name: 'Prerequisites'
+    permissions:
+      contents: read
+      pull-requests: read
+      statuses: read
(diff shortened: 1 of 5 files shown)


.github/workflows/perform-release.yaml
```diff
@@ -20,6 +20,12 @@ env:
 jobs:
   prerequisites:
     name: 'Prerequisites'
+    permissions:
+      contents: read
+      pull-requests: read
+      statuses: read
+      checks: read
```
marikaner, line 27: Not sure about those two. AI suggested it is needed.

Jonas-Isr (approved):

.github/workflows/perform-release.yaml
Jonas-Isr: Let's see what happens during next release :)

PR #844, opened 2026-04-28
feat: automated javadoc generation with artifact upload

n-o-u-r-h-a-n:
## Context

AI/ai-sdk-java-backlog#329.

Implemented a workflow which could be used for automatic javadoc generation by artifact upload. 
Here is the [link](https://github.com/SAP/ai-sdk-java/actions/runs/25060168553) to the successful continuous integration run where I tested the new workflow (javadoc.yaml) by calling it inside continuous integration.

### Feature scope:
 
- [x] Added the new workflow file named javadoc.yaml
- [x]  Integrated javadoc.yaml inside prepare-release workflow as well as perform release to create a PR with uploaded Aggregated JavaDoc artifact and update the JavaDocs on the documentation portal upon each version release of the SDK.

## Definition of Done

- [ ] ~[Documentation](https://github.com/SAP/ai-sdk/tree/main/docs-java) updated~
- [x] Release notes updated


.github/workflows/javadoc.yaml
```diff
@@ -0,0 +1,204 @@
...
+      - name: "Restore Dependencies"
+        uses: actions/cache/restore@v5
+        with:
+          key: ${{ env.MAVEN_CACHE_KEY }}
+          path: ${{ env.MAVEN_CACHE_DIR }}
+
+      - name: "Build SDK"
+        run: |
+          mvn -B -ntp -Dstyle.color=never -Drelease -DskipTests -Dgpg.skip=true \
+            clean install
```
Jonas-Isr: (Minor)

You can use `>` if there is only a single command that stretches over multiple lines to avoid the need to use `/`.

```suggestion
        run: >
          mvn -B -ntp -Dstyle.color=never -Drelease -DskipTests -Dgpg.skip=true
          clean install
```

.github/workflows/javadoc.yaml
```diff
@@ -0,0 +1,204 @@
...
+            delombok_dir="$module_path/target/delombok"
+            if [[ ! -d "$delombok_dir" ]]; then
+              echo "Missing delombok sources for $module_path" >&2
+              exit 1
+            fi
+            pkg_file="$module_path/target/javadoc-packages"
+            find "$delombok_dir" -name "*.java" \
+              | xargs grep -l "^package " \
+              | xargs grep "^package " \
+              | awk '{print $2}' | tr -d ';' | sort -u > "$pkg_file"
```
Jonas-Isr: (Major)

This one was found by Claude, so handle with care 😅 But I think this is actually a problem.

This will break silently if any file path contains spaces since `xargs` will split on them. It also runs `grep` twice (once to filter files, once to extract). If there are no .java files, `xargs` passes no arguments to `grep`, which then reads from `stdin` and hangs. Should use `xargs -r` (or `xargs --no-run-if-empty`) and `find ... -print0 | xargs -0`.

.github/workflows/javadoc.yaml
```diff
@@ -0,0 +1,204 @@
...
+          )
+
+          # Build classpath from all module jars + dependencies
+          classpath=""
+          for module_path in "${module_paths[@]}"; do
+            cp_file="$module_path/target/cp.txt"
+            mvn -B -ntp -f "$module_path/pom.xml" \
+              dependency:build-classpath -Dmdep.outputFile=target/cp.txt -Dmdep.includeScope=compile \
+              -Dgpg.skip=true -q
+            module_jar=$(find "$module_path/target" -maxdepth 1 -name "*.jar" ! -name "*sources*" ! -name "*javadoc*" | head -1)
```
Jonas-Isr, line 87: (Question)

As far as I see this only works reliably if there is only one .jar file, right? Not sure how relevant this is though...

.github/workflows/javadoc.yaml
```diff
@@ -0,0 +1,204 @@
...
+            local destdir="$2"
+            local pkg_file="$3"
+
+            if [[ ! -s "$pkg_file" ]]; then
+              echo "No packages found for $destdir, skipping" >&2
+              return
+            fi
+
+            mkdir -p "$destdir"
+            "$JAVADOC" \
```
Jonas-Isr: (Minor)

I think it would be a bit cleaner if you not simply rely on the `JAVADOC` variable defined outside of the function body here. Maybe add it as an additional parameter to the method?

.github/workflows/javadoc.yaml
```diff
@@ -0,0 +1,204 @@
...
+          mkdir -p "$out_dir"
+
+          module_paths=(
+            "core"
+            "orchestration"
+            "core-services/document-grounding"
+            "core-services/prompt-registry"
+            "foundation-models/openai"
+            "foundation-models/sap-rpt"
+          )
```
Jonas-Isr: (Minor)

Can we define`module_paths` top-level, e.g. in the `env` block (preferably together with the `module_names`)? That would have the upside that it is clearer what we have to change when we add the next module.

Jonas-Isr (commented):
The overall approach of using the bash script maybe is a very good alternative to using the plugins we had before. But I think to make it easier to maintain in the future we should add way more explanatory comments about what is actually happening.
Besides that I found some things worth checking but like the overall approach!

.github/workflows/javadoc.yaml
n-o-u-r-h-a-n: Yes, it only works when having one .jar file per module on the class path and if more than one .jar file is present, a one based on filesystem ordering will be picked up but that doesn't matter since any main JAR would have the same compiled .class files which Javadoc needs to understand the types.

.github/workflows/javadoc.yaml
n-o-u-r-h-a-n: Yes, sure I fixed it!

Jonas-Isr:
This looks good :) 

What is now missing is to call this workflow from the `prepare-release` workflow and to upload the generated files into our docs. For that we need
1) an additional step in your new workflow to create a PR on the [AI SDK docs repo](https://github.com/SAP/ai-sdk) with the newly created javadocs. This will be very close to what happens also in the [same workflow of the Cloud SDK](https://github.com/SAP/cloud-sdk-java/blob/033be690a99a3852a83af9230769d264b95a37c4/.github/workflows/javadoc.yaml#L109),
2) an additional step in the [prepare-release](https://github.com/SAP/ai-sdk-java/blob/main/.github/workflows/prepare-release.yaml) workflow to call your new workflow. This should be very similar to how this is called in the [same workflow of the Cloud SDK repo](https://github.com/SAP/cloud-sdk-java/blob/033be690a99a3852a83af9230769d264b95a37c4/.github/workflows/prepare-release.yaml#L157),
3) a note in the PR created in the [prepare-release](https://github.com/SAP/ai-sdk-java/blob/main/.github/workflows/prepare-release.yaml) workflow that the new javadoc PR needs to be reviewed and approved (exactly like [in Cloud SDK](https://github.com/SAP/cloud-sdk-java/blob/033be690a99a3852a83af9230769d264b95a37c4/.github/workflows/prepare-release.yaml#L308)),
4) two additional steps in the [perform release](https://github.com/SAP/ai-sdk-java/blob/main/.github/workflows/perform-release.yaml) workflow to merge the new javadoc PR. This should be very similar to how this is done in Cloud SDK [here](https://github.com/SAP/cloud-sdk-java/blob/033be690a99a3852a83af9230769d264b95a37c4/.github/workflows/perform-release.yml#L97) and [here](https://github.com/SAP/cloud-sdk-java/blob/033be690a99a3852a83af9230769d264b95a37c4/.github/workflows/perform-release.yml#L176).

This looks like a long list but it should be very close to what happens already in the Cloud SDK.

n-o-u-r-h-a-n:
> This looks good :)
> 
> What is now missing is to call this workflow from the `prepare-release` workflow and to upload the generated files into our docs. For that we need
> 
> 1. an additional step in your new workflow to create a PR on the [AI SDK docs repo](https://github.com/SAP/ai-sdk) with the newly created javadocs. This will be very close to what happens also in the [same workflow of the Cloud SDK](https://github.com/SAP/cloud-sdk-java/blob/033be690a99a3852a83af9230769d264b95a37c4/.github/workflows/javadoc.yaml#L109),
> 2. an additional step in the [prepare-release](https://github.com/SAP/ai-sdk-java/blob/main/.github/workflows/prepare-release.yaml) workflow to call your new workflow. This should be very similar to how this is called in the [same workflow of the Cloud SDK repo](https://github.com/SAP/cloud-sdk-java/blob/033be690a99a3852a83af9230769d264b95a37c4/.github/workflows/prepare-release.yaml#L157),
> 3. a note in the PR created in the [prepare-release](https://github.com/SAP/ai-sdk-java/blob/main/.github/workflows/prepare-release.yaml) workflow that the new javadoc PR needs to be reviewed and approved (exactly like [in Cloud SDK](https://github.com/SAP/cloud-sdk-java/blob/033be690a99a3852a83af9230769d264b95a37c4/.github/workflows/prepare-release.yaml#L308)),
> 4. two additional steps in the [perform release](https://github.com/SAP/ai-sdk-java/blob/main/.github/workflows/perform-release.yaml) workflow to merge the new javadoc PR. This should be very similar to how this is done in Cloud SDK [here](https://github.com/SAP/cloud-sdk-java/blob/033be690a99a3852a83af9230769d264b95a37c4/.github/workflows/perform-release.yml#L97) and [here](https://github.com/SAP/cloud-sdk-java/blob/033be690a99a3852a83af9230769d264b95a37c4/.github/workflows/perform-release.yml#L176).
> 
> This looks like a long list but it should be very close to what happens already in the Cloud SDK.

Do I need to update documentation and release notes?

.github/workflows/javadoc.yaml
```diff
@@ -0,0 +1,291 @@
...
+
+          cd .ai-sdk-docs
+          git add -A .
+
+          CHANGED_FILES="$(git status -s)"
+          if [[ -z "$CHANGED_FILES" ]]; then
+            echo "[DEBUG] No changes to API docs detected, skipping Pull Request creation."
+            echo "CREATE_PR=false" >> $GITHUB_OUTPUT
+            exit 0
+          fi
```
Jonas-Isr, line 278:
```suggestion
          CHANGED_FILES="$(git status -s)"
          if [[ -z "$CHANGED_FILES" ]]; then
            echo "[DEBUG] No changes to API docs detected, skipping Pull Request creation."
            exit 0
          fi
```

The `CREATE_PR` is never used.

.github/workflows/javadoc.yaml
```diff
@@ -0,0 +1,291 @@
...
+
+          BRANCH_NAME=java/release-docs-${{ steps.determine-version.outputs.CURRENT_VERSION }}
+          git switch --create "$BRANCH_NAME"
+          git commit -m "chore: Update JavaDocs for release ${{ steps.determine-version.outputs.CURRENT_VERSION }}"
+          git push origin "$BRANCH_NAME"
+
+          PR_TITLE="Java: Update JavaDocs for release ${{ steps.determine-version.outputs.CURRENT_VERSION }}"
+          PR_BODY="Replace the contents of v${{ steps.determine-version.outputs.MAJOR_VERSION }} API docs with the latest release of the SDK."
+          PR_URL=$(gh pr create --title "$PR_TITLE" --body "$PR_BODY" --repo "${{ env.DOCS_REPO }}")
+          echo "PR: $PR_URL" >> $GITHUB_STEP_SUMMARY
```
Jonas-Isr, line 290: (Minor)

Logging a bit might make debugging easier in the future.

```suggestion
          echo "Committing new JavaDocs"
          BRANCH_NAME=java/release-docs-${{ steps.determine-version.outputs.CURRENT_VERSION }}
          git switch --create "$BRANCH_NAME"
          git commit -m "chore: Update JavaDocs for release ${{ steps.determine-version.outputs.CURRENT_VERSION }}"
          git push origin "$BRANCH_NAME"

          echo "Creating PR"
          PR_TITLE="Java: Update JavaDocs for release ${{ steps.determine-version.outputs.CURRENT_VERSION }}"
          PR_BODY="Replace the contents of v${{ steps.determine-version.outputs.MAJOR_VERSION }} API docs with the latest release of the SDK."
          PR_URL=$(gh pr create --title "$PR_TITLE" --body "$PR_BODY" --repo "${{ env.DOCS_REPO }}")
          echo "PR: $PR_URL" >> $GITHUB_STEP_SUMMARY
```

Jonas-Isr:
> Do I need to update documentation and release notes?

Yes, good idea! How about a line like "- Aggregated JavaDocs are published now on our [documentation portal](https://sap.github.io/ai-sdk/docs/java/overview-cloud-sdk-for-ai-java)" under Improvements?

Jonas-Isr (commented):
Only very minor points!

n-o-u-r-h-a-n:
> > Do I need to update documentation and release notes?
> 
> Yes, good idea! How about a line like "- Aggregated JavaDocs are published now on our [documentation portal](https://sap.github.io/ai-sdk/docs/java/overview-cloud-sdk-for-ai-java)" under Improvements?

>Sounds nice! I added it!

Jonas-Isr (approved):
Nice!

PR #858, opened 2026-05-11
chore: [DevOps] Add `is-weekly` flag to spec-update workflow

Jonas-Isr:
## Context

Currently, the spec update workflows fail if there are no spec changes and the previous run using that same branch on the service's repo failed. This is intended behaviour for those workflow runs that are triggered from PRs on the services' repos.

This is however not the intended behaviour for those workflow runs that are triggered through the weekly spec update workflow.

An example of a workflow that should not have failed is [the spec update for document grounding from today](https://github.com/SAP/ai-sdk-java/actions/runs/25615445156/job/75192511022).

This PR introduces a flag to the spec update workflow that is set to true if the workflow is triggered from the weekly update workflow and bypasses the above described behaviour.

### DIFF
--- a/.github/workflows/spec-update.yaml
+++ b/.github/workflows/spec-update.yaml
@@ -35,6 +35,11 @@ on:
         required: false
         default: true
         type: boolean
+      is-weekly:
+        description: 'Flag to signal this is triggered by the weekly update run'
+        required: false
+        default: false
+        type: boolean
 
 env:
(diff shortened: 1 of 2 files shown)


CharlesDuboisSAP (commented):
LGTM, please do a test run from your branch

Jonas-Isr:
> LGTM, please do a test run from your branch

Good call!
You can find a succeeding run for grounding [here](https://github.com/SAP/ai-sdk-java/actions/runs/25724071658).

CharlesDuboisSAP (approved):

PR #869, opened 2026-05-15
fix: [DevOps] Run compile and test even if api compatibility test fails

rpanackal:
## Context

Previously, a failing compatibility skipped and marked all other tests as failing too.

image

---

**The new behaviour is** 

`Fail workflow later if API incompatible`, the `Compile and Test`, `Push changes` and `Create PR` steps should run before the fail step triggers.

### DIFF
--- a/.github/workflows/spec-update.yaml
+++ b/.github/workflows/spec-update.yaml
@@ -273,9 +273,6 @@ jobs:
             echo "api_compat_result=compatible" >> "$GITHUB_OUTPUT"
           else
             echo "api_compat_result=incompatible" >> "$GITHUB_OUTPUT"
-            echo "## ❌ API Compatibility: Breaking changes detected" >> $GITHUB_STEP_SUMMARY
-            echo "Download the **japicmp-report** artifact for details." >> $GITHUB_STEP_SUMMARY
-            exit 1
           fi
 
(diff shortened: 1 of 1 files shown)


.github/workflows/spec-update.yaml
```diff
@@ -386,6 +383,12 @@ jobs:
           echo "Client generation failed. Please check the Generate step logs for details."
           exit 1
 
+      - name: 'Fail if API incompatible'
+        if: steps.api_compat.outputs.api_compat_result == 'incompatible'
+        run: |
+          echo "API compatibility check detected breaking changes. Download the japicmp-report artifact for details."
```
Jonas-Isr: (Minor)

Where would we find the artifact? Is there a way to provide a useful link here for convenience?

rpanackal: You can find it on the workflow run summary. Here is an example run:

- https://github.com/SAP/ai-sdk-java/actions/runs/26347422785

image

Jonas-Isr:
```suggestion
          echo "API compatibility check detected breaking changes. Check the japicmp-report artifact in the workflow run summary for details."
```

CharlesDuboisSAP (approved):

PR #870, opened 2026-05-20
feat: PoC OpenAI Realtime API

n-o-u-r-h-a-n:
## Context

AI/ai-sdk-java-backlog#386.

Supported the Realtime API using the official OpenAI client.

### Feature scope:
 
- [x] Added and supported client secret key endpoints.
- [x] Added and supported 4 call service endpoints.
- [x] Added tests for newly supported endpoints.

## Definition of Done

- [ ] Release notes updated


n-o-u-r-h-a-n:
A clear understanding of what I've reached so far and implemented code accordingly:

The OpenAI Java SDK exposes all Realtime endpoints including SIP call control (accept, hangup, refer, reject) and client_secrets as plain HTTP POST requests, with no SIP protocol on the client side.

For example, when a phone call comes in, OpenAI's infrastructure hold the incoming SIP call open without answering it and notifies the backend with a call_id via webhook; the backend then calls `client.realtime().calls().accept(callId, params)`, which our SDK turns into a POST /v1/realtime/calls/{callId}/accept HTTP request forwarded through SAP Cloud SDK's Apache HTTP client (the same pipeline as the Responses API) and proxied by AI Core to OpenAI.   
 
OpenAI's infrastructure then connects the answered call to a Realtime AI session, and the actual WebRTC/SIP media flows directly between the user's device (browser/phone) and OpenAI.


What's supported now:

- [x] New `AiCoreOpenAiClient.realtime()` factory method exposing OpenAI SDK's RealtimeService
- [x] 5 realtime endpoints added to AiCoreHttpClientImpl allowlist
- [x] api-version=preview query param auto-injected by making AiCoreOpenAiClient use a dedicated `buildRealtimeClientOptions()` that adds the query parameter to ClientOptions, making the OpenAI SDK append it transparently to every realtime call.
- [x] Added 5 service methods in AiCoreOpenAiService.java.
- [x] WireMock unit tests for all 5 endpoints based on hardcoded json files.


Problems:

1. E2E test: `testCreateRealtimeClientSecret()` in AiCoreOpenAiTest.java  and the `/realtimeClientSecret` mapping in the index page when running the application are not working because there is no gpt-realtime deployment in the ai-sdk-java-e2e resource group on AI Core.

2. Tests in AiCoreOpenAiClientTest.java are depending on hardcoded json files regarding call id, new target uri of call transfer, etc.

vladimir-a-sap:
I'm closing this PR as we agreed in conversation in favor of https://github.com/SAP/ai-sdk-java/pull/935 due to the need of a public http hook endpoint with SIP approach, which is a problematic solution where our SDK may work.

If we decide to support SIP later we may reopen and update this PR

PR #875, opened 2026-05-27
chore: Remove `@Beta` annotations

CharlesDuboisSAP:
## Context

- https://github.com/SAP/cloud-sdk-java-backlog/issues/448
Remove `@Beta` annotations that accumulated over time

Previous PR for context:
- https://github.com/SAP/ai-sdk-java/pull/761

### DIFF
--- a/core-services/prompt-registry/src/main/java/com/sap/ai/sdk/prompt/registry/OrchestrationConfigClient.java
+++ b/core-services/prompt-registry/src/main/java/com/sap/ai/sdk/prompt/registry/OrchestrationConfigClient.java
@@ -4,7 +4,6 @@ import static com.sap.ai.sdk.core.JacksonConfiguration.getDefaultObjectMapper;
 
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
-import com.google.common.annotations.Beta;
 import com.sap.ai.sdk.core.AiCoreService;
 import com.sap.ai.sdk.prompt.registry.client.OrchestrationConfigsApi;
 import com.sap.ai.sdk.prompt.registry.model.AzureContentSafetyInputFilterConfig;
--- a/core-services/prompt-registry/src/main/java/com/sap/ai/sdk/prompt/registry/OrchestrationConfigClient.java
+++ b/core-services/prompt-registry/src/main/java/com/sap/ai/sdk/prompt/registry/OrchestrationConfigClient.java
@@ -4,6 +4,7 @@ import static com.sap.ai.sdk.core.JacksonConfiguration.getDefaultObjectMapper;
 
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.google.common.annotations.Beta;
 import com.sap.ai.sdk.core.AiCoreService;
 import com.sap.ai.sdk.prompt.registry.client.OrchestrationConfigsApi;
 import com.sap.ai.sdk.prompt.registry.model.AzureContentSafetyInputFilterConfig;
@@ -22,6 +23,7 @@ import lombok.val;
  *
  * @since 1.15.0
  */
+@Beta
 public class OrchestrationConfigClient extends OrchestrationConfigsApi {
 
   /**
--- a/core-services/prompt-registry/src/main/java/com/sap/ai/sdk/prompt/registry/OrchestrationConfigClient.java
+++ b/core-services/prompt-registry/src/main/java/com/sap/ai/sdk/prompt/registry/OrchestrationConfigClient.java
@@ -23,7 +23,6 @@ import lombok.val;
  *
  * @since 1.15.0
  */
-@Beta
 public class OrchestrationConfigClient extends OrchestrationConfigsApi {
 
   /**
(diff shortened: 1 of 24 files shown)


core-services/prompt-registry/src/main/java/com/sap/ai/sdk/prompt/registry/OrchestrationConfigClient.java
```diff
@@ -23,7 +22,6 @@
  *
  * @since 1.15.0
  */
-@Beta
 public class OrchestrationConfigClient extends OrchestrationConfigsApi {
```
rpanackal, line 27: (minor)

Keep this beta tag since it extends unstable (generated) code

core/src/main/java/com/sap/ai/sdk/core/common/ClientResponseHandler.java
```diff
@@ -27,7 +27,6 @@
  * @param <R> The type of the error response.
  * @since 1.1.0
  */
-@Beta
 @Slf4j
 @RequiredArgsConstructor
 public class ClientResponseHandler<T, R extends ClientError, E extends ClientException>
```
rpanackal, line 33: (minor/preferance)

I believe this was kept `@Beta` because it is public api for internal use only. I don't see the harm in keeping it since it discourages public use.

Same for `ClientStreamingHandler`, `RequestLogContext`. You can bring this to a larger round of discussion it slack.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/EmbeddingDeserializer.java
```diff
@@ -15,7 +14,6 @@
  * {@code @JsonCreator} annotations by intelligently detecting the JSON structure and choosing the
  * appropriate implementation.
  */
-@Beta
 class EmbeddingDeserializer extends JsonDeserializer<Embedding> {
```
rpanackal, line 19: (minor)

This extends Jackson API. Possibly breaking on migration to jackson 3

Same is true for `PolymorphicFallbackDeserializer`

core/src/main/java/com/sap/ai/sdk/core/common/ClientResponseHandler.java
rpanackal: You can ignore this since the alignment was to just have internal usage mentioned in javadoc.

core-services/prompt-registry/src/main/java/com/sap/ai/sdk/prompt/registry/OrchestrationConfigClient.java
CharlesDuboisSAP: We said generated code will not be annotated as Beta

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiBatchInput.java
```diff
@@ -38,6 +37,7 @@ public class OpenAiBatchInput {
    *
    * @param chatCompletionRequests the list of chat completion requests to include in the batch
    */
+  @Beta
```
Jonas-Isr, line 40: (Question)

What is the rational behind removing the tag on the class level but keeping it on this specific method?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationJacksonConfiguration.java
```diff
@@ -32,6 +31,7 @@ public class OrchestrationJacksonConfiguration {
    *     href="https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/converter/json/Jackson2ObjectMapperBuilder.html">Jackson2ObjectMapperBuilder</a>
    */
   @Nonnull
+  @Beta
```
Jonas-Isr, line 34: (Question)

Why do you move the tag here?

core-services/prompt-registry/src/main/java/com/sap/ai/sdk/prompt/registry/OrchestrationConfigClient.java
Jonas-Isr: > We said generated code will not be annotated as Beta

Yes, but this is not generated code and also for users does not look like generated code. So the instability of the API might not be obvious to users.

CharlesDuboisSAP: re-added annotation

foundation-models/openai/src/main/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiBatchInput.java
CharlesDuboisSAP: Because it was too general before, only the API with jackson should be annotated beta

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationJacksonConfiguration.java
CharlesDuboisSAP: Because it was too general before, only the API with jackson should be annotated beta

core-services/prompt-registry/src/main/java/com/sap/ai/sdk/prompt/registry/OrchestrationConfigClient.java
rpanackal: For future reference:

One other option is to actually stop extending generated code and follow the same approach as in other clients. That said I haven't looked too deeply at if such a change is too breaking.

Jonas-Isr (approved):
👍

Issue #888, opened 2026-06-04
DeploymentResolver selects deployments with status Stopped


Raphael-Kleindienst:
### Describe the Problem/Bug

For our Orchestrationclient configuration we followed the documentation [here](https://sap.github.io/ai-sdk/docs/java/connecting-to-ai-core#using-a-destination-from-the-btp-destination-service) to dynamically set the resource group for our OrchestrationClient.

This works, but if multiple orchestration deployments exist within the resource group then they are selected randomly or round-robin at run-time. In our case, one of the orchestration deployments had  "status": "STOPPED", leading to requests failing at random and with a misleading log-message ("Response from AI Core: Request failed with status 404 (Not Found): The response Content-Type is not JSON")

Please consider changing the way deployments are selected, so only active/running deployments are selected.

### Steps to Reproduce

1. Deploy multiple orchestration deployments to same resource group
2. Stop one of the deployments
3. Set up code as described

### Expected Behavior

Only running deployments are selected at runtime

### Used Versions

        <dependency>
            <groupId>com.sap.ai.sdk</groupId>
            <artifactId>core</artifactId>
            <version>1.17.0</version>
        </dependency>

### Code Examples

```kotlin
@Configuration
open class OrchestrationClientConfig(
    @Value("\${aicore.resource-group}") private val resourceGroup: String,
) {

    @Bean
    open fun orchestrationClient(): OrchestrationClient {
        val destination: HttpDestination  = AiCoreService().getInferenceDestination(resourceGroup).forScenario("orchestration")
        return OrchestrationClient(destination)
    }
}
```

### Log File

<details><summary>Log file</summary>
...
</details>

### Affected Development Phase

Production

### Impact

Inconvenience

rpanackal:
Thanks for pointing out the issue.

We do have some caching of deployments involved, which is currently unaware of the change of deployment status.

The original expectation was that the deployment will be maintained throughout runtime and there shouldn't be any behavioural difference the between choice of deployments. I can see that deployments stopped or deleted can be an issue.

I will try to reproduce the issue first and will get back to you soon.

Could you tell us a bit more about your use case? I am assuming you would have multiple deployments (atleast 2), when you stop one and then expect another to be picked up. Since deployments are effectively the same, why stop/delete in the first place?

Raphael-Kleindienst:
Hi @rpanackal,

[...]

PR #892, opened 2026-06-09
fix: [Core] Deployment resolution selects running targets

rpanackal:
## Context

- https://github.com/SAP/ai-sdk-java/issues/888

Deployment resolution at the moment picks the first deployment under given predicate. A predicate being a scenario, or model. This do not check for the status of the deployment. Therefore, if there are stopped deployments, it might be selected to construct the deployment url.

### Feature scope:
 
- [x] Insert a filter to only pick out deployments with target status running.

**Note:** Regular "status" points to the current status, but not the target state. "targetStatus" points to the expected state of the deployment avoiding the situation where a deployment get picked that will soon be obselete.

## Definition of Done

- [x] ~Release notes updated~

### DIFF
--- a/core/src/test/java/com/sap/ai/sdk/core/DeploymentResolverTest.java
+++ b/core/src/test/java/com/sap/ai/sdk/core/DeploymentResolverTest.java
@@ -179,7 +179,8 @@ class DeploymentResolverTest extends WireMockTestServer {
                     .withBodyFile("hasStoppedDeployment.json")
                     .withHeader("content-type", "application/json")));
 
-    final var deployment = resolver.getDeploymentIdByScenario(DEFAULT_RESOURCE_GROUP, "orchestration");
+    final var deployment =
+        resolver.getDeploymentIdByScenario(DEFAULT_RESOURCE_GROUP, "orchestration");
 
     assertThat(deployment).isEqualTo("d4b1396b84c1944d");
     assertThat(cache.get(DEFAULT_RESOURCE_GROUP))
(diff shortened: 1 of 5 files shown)


core/src/test/java/com/sap/ai/sdk/core/DeploymentResolverTest.java
```diff
@@ -169,6 +170,26 @@ void testIsDeploymentOfModel() {
...
+
+    final var deployment =
+        resolver.getDeploymentIdByScenario(DEFAULT_RESOURCE_GROUP, "orchestration");
+
+    assertThat(deployment).isEqualTo("d4b1396b84c1944d");
+    assertThat(cache.get(DEFAULT_RESOURCE_GROUP))
+        .extracting(AiDeployment::getId, AiDeployment::getTargetStatus)
+        .contains(
+            tuple("d2a491b5010620b0", AiDeployment.TargetStatusEnum.STOPPED),
+            tuple("d4b1396b84c1944d", AiDeployment.TargetStatusEnum.RUNNING));
```
CharlesDuboisSAP: Why yould the cache contain the `STOPPED` deployment?

rpanackal: Because the cache loads all deployments as returned by Deployments API. Currently, the cache is for all deployments under a resource group.

But you are right. We can load the cache selectively instead. 

Essentially, instead of caching all deployments, we could only cache `RUNNING` deployments

rpanackal: Is that preferrable to you? Do you see any logic that relies on the cache being an exhaustive list of deployments that might break?

I am a bit skeptical about touching a global cache,

CharlesDuboisSAP: No reason to cache deployments that can't be used

CharlesDuboisSAP (approved):

CharlesDuboisSAP (changes requested):
Release note?

CharlesDuboisSAP (approved):

PR #895, opened 2026-06-10
chore: Fix CVEs and update Cloud SDK version

Jonas-Isr:
## Context

This PR updates the Cloud SDK version and fixes version to fix the CVEs mentioned [here](https://github.com/SAP/ai-sdk-java/actions/runs/27254601571/job/80486091402#step:8:195)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,7 @@
     <maven.compiler.proc>full</maven.compiler.proc>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.build.outputTimestamp>2025-04-03T13:23:00Z</project.build.outputTimestamp>
-    <cloud-sdk.version>5.30.0</cloud-sdk.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -103,6 +104,13 @@
 
   <dependencyManagement>
     <dependencies>
+      <dependency>
+        <groupId>org.springframework</groupId>
+        <artifactId>spring-framework-bom</artifactId>
+        <version>${spring-framework.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
```
CharlesDuboisSAP: I would rather not have springframework in the parent pom

Jonas-Isr: Yes, that was ugly but I did not find a better way before (because Spring AI carried the vulnerable Spring framework version as well).

But this is all unnecessary now since there is a new Spring AI version that removes the vulnerabilities.

CharlesDuboisSAP (approved):

PR #921, opened 2026-06-30
fix: [DevOps] Temporary for NVD, use datafeed instead of API

CharlesDuboisSAP:
## Context

AI/ai-sdk-java-backlog#448.

If the current dependency vulnerability scan fails, then we move temporarily to a data feed, higher bandwidth but higher reliability.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
     <javaparser.version>3.28.2</javaparser.version>
     <jsonschema-generator.version>4.38.0</jsonschema-generator.version>
     <jackson.version>2.22.0</jackson.version>
-    <logback.version>1.5.36</logback.version>
+    <logback.version>1.5.37</logback.version>
     <commons-codec.version>1.22.0</commons-codec.version>
     <japicmp.version>0.26.1</japicmp.version>
     <!-- conflicts resolution -->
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -828,6 +828,7 @@ https://gitbox.apache.org/repos/asf?p=maven-pmd-plugin.git;a=blob_plain;f=src/ma
         <version>12.1.0</version>
         <configuration>
           <connectionTimeout>60000</connectionTimeout>
+          <nvdDatafeedUrl>https://dependency-check.github.io/DependencyCheck_Builder/nvd_cache/nvdcve-{0}.json.gz</nvdDatafeedUrl>
```
Jonas-Isr, line 831: (Question)

This automatically jumps in if the call to the usual server fails?

Jonas-Isr (commented):
Can we just run it from the branch and see if it works?

CharlesDuboisSAP:
> Can we just run it from the branch and see if it works?

It works (ish) https://github.com/SAP/ai-sdk-java/actions/runs/28446859432/job/84298387124

pom.xml
CharlesDuboisSAP: It replaces the server

Jonas-Isr:
> It works (ish) https://github.com/SAP/ai-sdk-java/actions/runs/28446859432/job/84298387124

Why the "ish"? It looks like it is working properly right? Or do you mean because it found a violation?

CharlesDuboisSAP:
> Why the "ish"? It looks like it is working properly right? Or do you mean because it found a violation?

[I just fixed the violation](https://github.com/SAP/ai-sdk-java/actions/runs/28518789246)

Jonas-Isr (approved):
Looks good! We just need to remember to change back to the old version at some point.

PR #936, opened 2026-07-08
chore: [OpenAI] Improve e2e test stability

Jonas-Isr:
## Context

Changed the `WeatherMethod` example used for tool calls:
- removed the `Request` inner class,
- instead use two primitive arguments.

Also changed all uses of the test class accordingly

Please also consider the docs PR:
- https://github.com/SAP/ai-sdk/pull/580

### DIFF
--- a/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OpenAiService.java
+++ b/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OpenAiService.java
@@ -120,14 +120,13 @@ public class OpenAiService {
     final var messages = new ArrayList<OpenAiMessage>();
     messages.add(OpenAiMessage.user("What's the weather in %s in %s?".formatted(location, unit)));
 
-    record WeatherRequest(String location, WeatherMethod.Unit unit) {}
-
     // 1. Define the function
     final List<OpenAiTool> tools =
         List.of(
             OpenAiTool.forFunction(
-                    (WeatherRequest r) -> WeatherMethod.getCurrentWeather(r.location(), r.unit()))
-                .withArgument(WeatherRequest.class)
+                    (WeatherMethod.Request r) ->
+                        WeatherMethod.getCurrentWeather(r.location(), r.unit()))
+                .withArgument(WeatherMethod.Request.class)
                 .withName("weather")
                 .withDescription("Get the weather for the given location"));
 
--- a/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OpenAiService.java
+++ b/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OpenAiService.java
@@ -120,11 +120,14 @@ public class OpenAiService {
     final var messages = new ArrayList<OpenAiMessage>();
     messages.add(OpenAiMessage.user("What's the weather in %s in %s?".formatted(location, unit)));
 
+    record WeatherRequest(String location, WeatherMethod.Unit unit) {}
+
     // 1. Define the function
     final List<OpenAiTool> tools =
         List.of(
-            OpenAiTool.forFunction(WeatherMethod::getCurrentWeather)
-                .withArgument(WeatherMethod.Request.class)
+            OpenAiTool.forFunction(
+                    (WeatherRequest r) -> WeatherMethod.getCurrentWeather(r.location(), r.unit()))
+                .withArgument(WeatherRequest.class)
                 .withName("weather")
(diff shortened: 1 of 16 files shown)


sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OpenAiService.java
```diff
@@ -120,11 +120,14 @@ public OpenAiChatCompletionResponse chatCompletionToolExecution(
     final var messages = new ArrayList<OpenAiMessage>();
     messages.add(OpenAiMessage.user("What's the weather in %s in %s?".formatted(location, unit)));
 
+    record WeatherRequest(String location, WeatherMethod.Unit unit) {}
+
     // 1. Define the function
     final List<OpenAiTool> tools =
         List.of(
-            OpenAiTool.forFunction(WeatherMethod::getCurrentWeather)
-                .withArgument(WeatherMethod.Request.class)
```
CharlesDuboisSAP: Why did you move the Request from the `WeatherMethod.Request` to the service class, it's duplication.

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/SpringAiOpenAiTest.java
```diff
@@ -76,9 +76,9 @@ void testToolCallingWithoutExecution() {
     assertThat(toolCall1.name()).isEqualTo("getCurrentWeather");
     assertThat(toolCall2.name()).isEqualTo("getCurrentWeather");
     assertThat(toolCall1.arguments())
-        .matches("\\{\"arg0\": \\{\"location\": \"[^\"]*Potsdam[^\"]*\", \"unit\": \"C\"}}");
+        .matches("\\{\"arg0\": \"[^\"]*Potsdam[^\"]*\", \"arg1\": \"C\"}");
```
CharlesDuboisSAP, line 79: Can't arguments be properly parsed by spring to as `location` and `unit`? `arg0` might be worse for reliability

CharlesDuboisSAP (commented):
Should we also change the unit tests and documentation as well?

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OpenAiService.java
Jonas-Isr: You need this workaround since `forFunction()` in `OpenAiTool` needs a function with only a single argument as input. To then still use the `WeatherMethod.getCurrentWeather()` method you need the local WeatherRequest method. But this ensures that the `WeatherMethod.getCurrentWeather()` is only called with primitive arguments.

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/SpringAiOpenAiTest.java
Jonas-Isr: As far as I know this is the (only) way Spring does it.

Jonas-Isr:
> Should we also change the unit tests and documentation as well?

Yes, changed the unit tests as well. For the documentation see the PR linked in the description above.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OpenAiService.java
CharlesDuboisSAP: That doesn't answer the question, I didn't talk about `getCurrentWeather `?

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/SpringAiOpenAiTest.java
CharlesDuboisSAP: I had managed to do it with Copilot a couple weeks ago but reverted it, I think it's possible

Jonas-Isr: Decision after discussion: we leave as is for now and reconsider if it keeps being flaky.

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OpenAiService.java
Jonas-Isr: Fixed

CharlesDuboisSAP (approved):
This might not be reliable because it still sends arg0 and arg1, let's see

PR #939, opened 2026-07-09
feat: [Orchestration] reasoning content 

n-o-u-r-h-a-n:
## Context

AI/ai-sdk-java-backlog#412.

As a user, I would like to see the thinking processes when using the reasoning models, in addition to the final outcome. This is relevant for synchronous and streaming use-cases.

The harmonized API of orchestration supports this feature (see link above), so the SDK should also follow.

### Feature scope:
 
- [x] supported reasoning content using model classes code
- [x] added sample code as well as updated controller and index page  
- [x] added e2e and unit tests  

## Definition of Done

- [x] Release notes updated


orchestration/src/main/java/com/sap/ai/sdk/orchestration/AssistantMessage.java
```diff
@@ -61,12 +67,16 @@ public AssistantMessage(@Nonnull final String singleMessage) {
...
 
   private AssistantMessage(
-      @Nonnull final MessageContent content, @Nullable final List<MessageToolCall> toolCalls) {
+      @Nonnull final MessageContent content,
+      @Nullable final List<MessageToolCall> toolCalls,
+      @Nullable final List<ReasoningBlock> reasoningContent) {
     this.content = content;
     this.toolCalls = toolCalls;
+    this.reasoningContent = reasoningContent;
   }
```
CharlesDuboisSAP: Can this be replaced by lombok's `@RequiredArgsContructor`

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AssistantMessage.java
```diff
@@ -81,7 +91,22 @@ private AssistantMessage(
...
+   * @param reasoningContent the reasoning blocks from the previous assistant turn.
+   * @return a new AssistantMessage instance with the reasoning content set.
+   * @see <a href="https://help.sap.com/docs/sap-ai-core/generative-ai/reasoning">SAP AI Core:
+   *     Orchestration - Reasoning</a>
+   */
+  @SuppressWarnings("PMD.PublicApiExposesModelType")
+  @Nonnull
+  public AssistantMessage withReasoningContent(
+      @Nonnull final List<ReasoningBlock> reasoningContent) {
+    return new AssistantMessage(this.content, this.toolCalls, List.copyOf(reasoningContent));
```
CharlesDuboisSAP, line 100: You should consider that the warning is there for a reason, You could return a list of string instead or make your own convenience class.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
```diff
@@ -131,6 +132,54 @@ public Stream<String> streamChatCompletion(
...
+   *     Orchestration - Reasoning</a>
+   */
+  @SuppressWarnings("PMD.PublicApiExposesModelType")
+  @Nonnull
+  public List<ReasoningBlock> streamReasoning(
+      @Nonnull final OrchestrationPrompt prompt,
+      @Nonnull final OrchestrationModuleConfig config,
+      @Nonnull final Consumer<String> onAnswerChunk,
+      @Nonnull final Consumer<String> onReasoningChunk,
+      @Nonnull final OrchestrationModuleConfig... fallbackConfigs)
```
CharlesDuboisSAP: Not sure how I feel about those 2 new public methods that only add a `forEach` on an existing public `stream` method.
Reasoning is a very specific feature and if we add public API for every single feature we would have way to many methods.
Also the customers could add the `forEach` themselves and it would allow more configurability.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AssistantMessage.java
n-o-u-r-h-a-n: There was @Value annotation which already creates an AllArgsConstructor so I added a line to make its access level private.

CharlesDuboisSAP: You can add a parameter in `@RequiredArgsContructor` to make it private instead

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AssistantMessage.java
ZhongpinWang: This `@Getter` is redundant due to `@Value` at the class level

ZhongpinWang: This `@Getter` is redundant as `@Value` already creates Getter for all fields.

Quote from lombok doc:

> `In practice, @Value is shorthand for: final @ToString @EqualsAndHashCode @AllArgsConstructor @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE) @Getter`

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AssistantMessage.java
n-o-u-r-h-a-n: Okay, I found this way which I just pushed which creates ReasoningItem which is a hand-written convenience record (String content,String signature) that wraps the generated ReasoningBlock model type preventing the auto generated OpenAI class from leaking into SDK's public surface.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
n-o-u-r-h-a-n: image

This would be the alternative solution which users will do if we decided to remove the public API.

ZhongpinWang (commented):
@CharlesDuboisSAP Regarding the signature, it does not need a convenience to get it out. It only needs to be as part of the message history and be sent back together in the next LLM call for verification reasons. So IMO providing a convenience to access reasoning content text would be enough.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AssistantMessage.java
CharlesDuboisSAP: [Zhongpin mentioned](https://github.com/SAP/ai-sdk-java/pull/939#pullrequestreview-4662618061) that you don’t need a convenience for the reasoning signature, you can directly return string

Jonas-Isr: Just as an explanation:
The `@SuppressWarnings("PMD.PublicApiExposesModelType")` you see sometimes in our code is only on methods that we made public before we decided that we do not want to expose generated model types in our convenience API. By now, we added this PMD rule and should stick to it in any new code. But since we cannot change the old public API (since it would break code users of the SDK might rely on), we needed the suppression tag for the old code.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationClient.java
Jonas-Isr: Is this code snippet from the service class? 

While I see your point that this is a lot of code for users to write, I would also agree with Charles that putting this into a convenience method is maybe too much. The reason for this is that I think the level of complexity the users have to understand (5 different inputs in the method, why 2 different consumers, asynchronicity,...) compared to the fact that we do not know how useful this actually will be for users (maybe they don't want to be streaming reasoning in this way or want to do something with it before parsing it to the consumer) does not match up. Usability is probably something we could argue over but the current complexity of using this convenience method I think is not "convenient" enough. 

Plus, we need to publish a lot of new API for this (whole new public class `ReasoningAccumulator` that we otherwise don't need as far as I see). Since we would then forever need to kepp mantaining this I think it is not worth it.

n-o-u-r-h-a-n: Yes, I agree

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OrchestrationService.java
```diff
@@ -179,6 +183,95 @@ public Stream<String> streamChatCompletion(@Nonnull final String topic) {
...
+   * @param topic the topic to ask about.
+   * @param onAnswerChunk consumer invoked with each answer text chunk.
+   * @param onReasoningChunk consumer invoked with each reasoning text chunk.
+   * @return the assembled reasoning text as returned by the API.
+   */
+  @Nonnull
+  public List<String> streamReasoning(
+      @Nonnull final String topic,
+      @Nonnull final Consumer<String> onAnswerChunk,
+      @Nonnull final Consumer<String> onReasoningChunk) {
```
CharlesDuboisSAP: Why don't you return two streams in an object, which then can get consumed without an accumulator. It would be simpler and more comparable to existing streaming endpoints

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AssistantMessage.java
```diff
@@ -6,31 +6,40 @@
...
 
   /** The content of the message. */
-  @Nonnull @Getter MessageContent content;
+  @Nonnull MessageContent content;
 
   /** Tool call if there is any. */
   @Nullable List<MessageToolCall> toolCalls;
 
+  /** Reasoning (thinking) blocks from the model's previous turn. */
+  @SuppressWarnings("PMD.PublicApiExposesModelType")
```
Jonas-Isr: (Minor)

I think this can be deleted as it is not necessary anymore.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationChatResponse.java
```diff
@@ -67,6 +68,30 @@ public TokenUsage getTokenUsage() {
...
+
+  /**
+   * Get the reasoning (thinking) content as a single joined string, for display purposes.
+   *
+   * @return the concatenated reasoning text; empty if there is no reasoning.
+   */
+  @Nonnull
+  public String getReasoningText() {
+    return String.join("", getReasoningContent());
+  }
```
Jonas-Isr: (Minor)

Maybe it would be nice to have an overload if users want a different delimiter: `public String getReasoningText(String delimiter)`?

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/OrchestrationTest.java
```diff
@@ -688,4 +690,45 @@ void testCitations() {
     val result = service.citations();
     assertThat(result.getOriginalResponse().getFinalResult().getCitations()).isNotEmpty();
   }
+
+  @Test
+  void testReasoningContentSync() {
+    val result = service.reasoning("What is 6 times 7?");
+
+    assertThat(result.answer()).isNotEmpty();
+    assertThat(result.reasoning()).isNotNull();
```
Jonas-Isr: The list will always be nonnull.

```suggestion
    assertThat(result.reasoning()).isNotEmpyt();
```

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/OrchestrationTest.java
```diff
@@ -688,4 +690,45 @@ void testCitations() {
...
+  @Test
+  void testMultiTurnReasoning() {
+    val reasoningConfig =
+        new OrchestrationModuleConfig()
+            .withLlmConfig(CLAUDE_4_5_SONNET.withReasoningEffort(ReasoningEffort.MEDIUM));
+    val orchestrationClient = new OrchestrationClient();
+    val firstPrompt = new OrchestrationPrompt("What is 6 times 7?");
+    val firstResponse = orchestrationClient.chatCompletion(firstPrompt, reasoningConfig);
+
+    assertThat(firstResponse.getReasoningContent()).isNotNull();
```
Jonas-Isr: See above.

```suggestion
    assertThat(firstResponse.getReasoningContent()).isNotEmpty();
```

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/OrchestrationTest.java
```diff
@@ -688,4 +690,45 @@ void testCitations() {
...
+    // Feed the whole history (which carries the previous reasoning content internally) into the
+    // follow-up turn.
+    val followUpPrompt =
+        new OrchestrationPrompt("Are you sure?").messageHistory(firstResponse.getAllMessages());
+    val followUp = orchestrationClient.chatCompletion(followUpPrompt, reasoningConfig);
+
+    assertThat(followUp.getContent()).isNotEmpty();
+    // Reasoning is still returned on the follow-up turn, proving the model kept reasoning enabled
+    // after we resubmitted the previous reasoning_content in messages_history.
+    assertThat(followUp.getReasoningContent()).isNotNull();
```
Jonas-Isr: See above.


```suggestion
    assertThat(followUp.getReasoningContent()).isNotEmpty();
```

orchestration/src/test/resources/reasoningRequest.json
```diff
@@ -0,0 +1,29 @@
+{
```
Jonas-Isr, line 1: (Question)

Maybe I am mistaken but I think this response file and `streamReasoningRequest.json` are never actually used. At least I do not see it int he tests were the response is checked.
(You do this in OrchestrationUnitTest.multiTurnReasoningPreservesReasoningContent()`, but I don't see it for these two.)

orchestration/src/main/java/com/sap/ai/sdk/orchestration/AssistantMessage.java
n-o-u-r-h-a-n: Yes, forgot to remove it after fixing that users won't see ReasoningBlock anymore.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationChatResponse.java
n-o-u-r-h-a-n: Yes, works even better! I added it.

orchestration/src/test/resources/reasoningRequest.json
n-o-u-r-h-a-n: Yes, I'm adding extra wiring for them in the tests now.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ReasoningAccumulator.java
```diff
@@ -0,0 +1,56 @@
...
+import javax.annotation.Nonnull;
+
+/**
+ * Accumulates reasoning ({@code reasoning_content}) text across the chunks of a streaming
+ * orchestration response, merging the pieces of each block by their position.
+ *
+ * @see <a href="https://help.sap.com/docs/sap-ai-core/generative-ai/reasoning">SAP AI Core:
+ *     Orchestration - Reasoning</a>
+ */
+public class ReasoningAccumulator {
```
Jonas-Isr: (Personal Opinion)

I am not sure if we need this class. I think users should write these themselves.

ZhongpinWang: PS: JS has accumulator, mainly for aggregating all streaming properties, so that at the end of the streaming process, users can get an object similar to (or I think even same as) the non-streaming response object. But this is just optional :)

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/OrchestrationService.java
n-o-u-r-h-a-n: > Why don't you return two streams in an object, which then can get consumed without an accumulator. It would be simpler and more comparable to existing streaming endpoints

I fixed it by returning one stream: Stream<ReasoningChunk> where ReasoningChunk is `public record ReasoningOutput(@Nonnull String answer, @Nonnull List<String> reasoning) {}`.

Accumulator is removed entirely.

Tell me how you feel about it now?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ReasoningAccumulator.java
n-o-u-r-h-a-n: I also added the following alternative solution to the accumulator as described in [this somment](https://github.com/SAP/ai-sdk-java/pull/939#discussion_r3580530768).

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
```diff
@@ -79,6 +79,48 @@ ResponseEntity<ResponseBodyEmitter> streamChatCompletion() {
...
+                  for (final var reasoning : chunk.reasoning()) {
+                    if (!reasoning.isEmpty()) {
+                      send(emitter, "[reasoning] " + reasoning);
+                    }
+                  }
+                });
+          } finally {
+            emitter.complete();
+          }
+        };
```
CharlesDuboisSAP, line 161:
```suggestion
          final var lastLabel = new String[] {""};

          try (var stream = service.streamReasoning(topic)) {
            stream.forEach(
                chunk -> {
                  if (!chunk.answer().isEmpty()) {
                    if (!"answer".equals(lastLabel[0])) {
                      send(emitter, "\n-----ANSWER-----\n");
                      lastLabel[0] = "answer";
                    }
                    send(emitter, chunk.answer());
                  }
                  for (final var reasoning : chunk.reasoning()) {
                    if (!reasoning.isEmpty()) {
                      if (!"reasoning".equals(lastLabel[0])) {
                        send(emitter, "\n----REASONING----\n");
                        lastLabel[0] = "reasoning";
                      }
                      send(emitter, reasoning);
                    }
                  }
                });
```

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
```diff
@@ -79,6 +79,48 @@ ResponseEntity<ResponseBodyEmitter> streamChatCompletion() {
...
+  Object reasoning(
+      @RequestParam(value = "topic", defaultValue = "Why is the sky blue?") final String topic) {
+    return service.reasoning(topic);
+  }
+
+  @GetMapping("/multiTurnReasoning")
+  Object multiTurnReasoning(
+      @RequestParam(value = "first", defaultValue = "What is 6 times 7?") final String first,
+      @RequestParam(value = "followUp", defaultValue = "Are you sure?") final String followUp) {
+    return service.multiTurnReasoning(first, followUp);
```
CharlesDuboisSAP: Formatting between message and JSON object is missing for both endpoints

sample-code/spring-app/src/main/resources/static/index.html
```diff
@@ -266,6 +266,42 @@ <h2>Orchestration</h2>
...
                                 </li>
+                                <li class="list-group-item">
+                                    <div class="info-tooltip">
+                                        <button type="submit"
+                                                formaction="/orchestration/reasoning"
+                                                class="link-offset-2-hover link-underline link-underline-opacity-0 link-underline-opacity-75-hover endpoint">
+                                            <code>/orchestration/reasoning</code>
+                                        </button>
+                                        <div class="tooltip-content">
+                                            Chat request against a reasoning-capable model. Returns both the final answer and the reasoning blocks.
```
CharlesDuboisSAP:
```suggestion
                                            Chat request against a reasoning-capable model. Returns both the final answer and its reasoning.
```

sample-code/spring-app/src/main/resources/static/index.html
```diff
@@ -266,6 +266,42 @@ <h2>Orchestration</h2>
...
+                                </li>
+                                <li class="list-group-item">
+                                    <div class="info-tooltip">
+                                        <button type="submit"
+                                                formaction="/orchestration/streamReasoning"
+                                                class="link-offset-2-hover link-underline link-underline-opacity-0 link-underline-opacity-75-hover endpoint">
+                                            <code>/orchestration/streamReasoning</code>
+                                        </button>
+                                        <div class="tooltip-content">
+                                            Streaming variant that emits answer and reasoning chunks live, followed by the accumulated reasoning blocks.
```
CharlesDuboisSAP:
```suggestion
                                            Streaming variant that both answer and reasoning chunks.
```

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/OrchestrationTest.java
```diff
@@ -688,4 +691,51 @@ void testCitations() {
...
+    try (val stream = service.streamReasoning("What is 6 times 7?")) {
+      stream.forEach(
+          chunk -> {
+            answerBuilder.append(chunk.answer());
+            reasoning.addAll(chunk.reasoning());
+          });
+    }
+
+    assertThat(answerBuilder.toString()).isNotEmpty();
+    assertThat(reasoning).isNotNull();
```
CharlesDuboisSAP: reasoning is never null, this is a bad assertion

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationChatCompletionDelta.java
```diff
@@ -27,6 +29,24 @@ public String getDeltaContent() {
...
+   *
+   * @return the reasoning text of each block in this chunk; never {@code null}, may be empty.
+   * @see <a href="https://help.sap.com/docs/sap-ai-core/generative-ai/reasoning">SAP AI Core:
+   *     Orchestration - Reasoning</a>
+   */
+  @Nonnull
+  public List<String> getDeltaReasoningContent() {
+    val choices = getFinalResult().getChoices();
+    if (!choices.isEmpty() && choices.get(0).getIndex() == 0) {
+      return choices.get(0).getDelta().getReasoningContent().stream()
```
CharlesDuboisSAP: Could you ask someone in the orchestration team why is `reasoning_content` a list? If there is more than one element is it the same reasoning sentence or a parallel one that shouldn't be appended? If it's multiple elements of the same sentence we could directly return a string instead of this inconvenient list.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationChatResponse.java
```diff
@@ -67,6 +68,44 @@ public TokenUsage getTokenUsage() {
...
+   *
+   * @return the reasoning text of each block; never {@code null}, may be empty.
+   * @see <a href="https://help.sap.com/docs/sap-ai-core/generative-ai/reasoning">SAP AI Core:
+   *     Orchestration - Reasoning</a>
+   */
+  @Nonnull
+  public List<String> getReasoningContent() {
+    return getChoice().getMessage().getReasoningContent().stream()
+        .map(ReasoningBlock::getContent)
+        .toList();
```
CharlesDuboisSAP: Same comment, why does function that returns list exists?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ReasoningEffort.java
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Wire string sent as the value of the {@code reasoning_effort} model parameter.
+   *
+   * @return the wire value.
+   */
+  @Nonnull
+  public String getValue() {
+    return value;
+  }
```
CharlesDuboisSAP: Can be replaced by lombok

sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/controllers/OrchestrationController.java
```diff
@@ -79,6 +79,54 @@ ResponseEntity<ResponseBodyEmitter> streamChatCompletion() {
...
+                  for (final var reasoning : chunk.reasoning()) {
+                    if (!reasoning.isEmpty()) {
+                      if (!"reasoning".equals(lastLabel[0])) {
+                        send(emitter, "\n----REASONING----\n");
+                        lastLabel[0] = "reasoning";
+                      }
+                      send(emitter, reasoning);
+                    }
+                  }
+                });
```
CharlesDuboisSAP, line 157:
```suggestion
                });
          } finally {
            emitter.complete();
          }
        };
```
My bad I removed it by accident

n-o-u-r-h-a-n: no problem, I added it

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationChatCompletionDelta.java
n-o-u-r-h-a-n: Agreed to release it with only the string and when we know of a use-case for the list we add it later.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationChatResponse.java
```diff
@@ -100,10 +87,14 @@ public String getReasoningText() {
...
    * @return the joined reasoning text; empty if there is no reasoning.
+   * @see <a href="https://help.sap.com/docs/sap-ai-core/generative-ai/reasoning">SAP AI Core:
+   *     Orchestration - Reasoning</a>
    */
   @Nonnull
   public String getReasoningText(@Nonnull final String delimiter) {
-    return String.join(delimiter, getReasoningContent());
+    return getChoice().getMessage().getReasoningContent().stream()
+        .map(ReasoningBlock::getContent)
+        .collect(Collectors.joining(delimiter));
```
CharlesDuboisSAP: If there is only one block, there is no need for a function that join blocks separated by a delimiter

orchestration/src/test/java/com/sap/ai/sdk/orchestration/OrchestrationReasoningContentTest.java
```diff
@@ -74,7 +74,6 @@ void assistantMessageSerializesReasoningContentVerbatim() {
   void chatResponseGetReasoningContentReturnsEmptyWhenAbsent() throws Exception {
     val response = parseSyncResponse("");
 
-    assertThat(response.getReasoningContent()).isEmpty();
     assertThat(response.getReasoningText()).isEmpty();
```
CharlesDuboisSAP, line 77: Why did you remove the `getReasoningContent` assertion lines 77 and 87?

CharlesDuboisSAP (commented):
Minor comments, looks good

orchestration/src/test/java/com/sap/ai/sdk/orchestration/OrchestrationReasoningContentTest.java
n-o-u-r-h-a-n: because I removed the getReasoningContent() method and kept only the getReasoningText one which returns a string.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationChatResponse.java
n-o-u-r-h-a-n: yes, fixed

orchestration/src/main/java/com/sap/ai/sdk/orchestration/OrchestrationChatResponse.java
```diff
@@ -67,6 +69,21 @@ public TokenUsage getTokenUsage() {
...
+  /**
+   * Get the reasoning (thinking) content produced by the model, as text, for display purposes. When
+   * the model returns multiple reasoning blocks, they are joined with a single space.
+   *
+   * @return the reasoning text; empty if there is no reasoning.
+   * @see <a href="https://help.sap.com/docs/sap-ai-core/generative-ai/reasoning">SAP AI Core:
+   *     Orchestration - Reasoning</a>
+   */
+  @Nonnull
+  public String getReasoningText() {
```
CharlesDuboisSAP, line 81: (Minor) `getReasoningText` is not consistent with `getDeltaReasoningContent` (text and content)

n-o-u-r-h-a-n: I renamed it to `getDeltaReasoningText()`

CharlesDuboisSAP (approved):

PR #944, opened 2026-07-14
chore: [DevOps] Clean up logs

Jonas-Isr:
## Context

- Tried to originally fix the e2e test, but https://github.com/SAP/ai-sdk-java/pull/950 fixed it instead. This is still a nice improvement, removing logs to clean up

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -55,7 +55,9 @@ jobs:
           awk '/Results:/, /----/' mvn_output.log > test_error.log || true # true ensures the step doesn't fail if no match is found.
           ERROR_MSG=$(cat test_error.log | tail +4 | head -n -4 | awk 'BEGIN{RS=""} {gsub(/"/, "\\\""); gsub(/Expecting KeySet:.*but could not find the following element\(s\):/, "Missing element: "); printf "%s\n", $0}')
           echo "$ERROR_MSG"
-          echo "error_message=$ERROR_MSG" >> $GITHUB_OUTPUT
+          echo "error_message<<GITHUB_OUTPUT_EOF" >> $GITHUB_OUTPUT
+          echo "$ERROR_MSG" >> $GITHUB_OUTPUT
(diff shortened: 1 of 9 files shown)


.github/workflows/e2e-test.yaml
```diff
@@ -55,7 +55,9 @@ jobs:
           awk '/Results:/, /----/' mvn_output.log > test_error.log || true # true ensures the step doesn't fail if no match is found.
           ERROR_MSG=$(cat test_error.log | tail +4 | head -n -4 | awk 'BEGIN{RS=""} {gsub(/"/, "\\\""); gsub(/Expecting KeySet:.*but could not find the following element\(s\):/, "Missing element: "); printf "%s\n", $0}')
           echo "$ERROR_MSG"
-          echo "error_message=$ERROR_MSG" >> $GITHUB_OUTPUT
+          echo "error_message<<GITHUB_OUTPUT_EOF" >> $GITHUB_OUTPUT
+          echo "$ERROR_MSG" >> $GITHUB_OUTPUT
+          echo "GITHUB_OUTPUT_EOF" >> $GITHUB_OUTPUT
```
CharlesDuboisSAP: What purpose does putting message<<EOF in the output have

Jonas-Isr: It's documented [here](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#multiline-strings).

Basically, the first line tells github to treat everything that comes next until `GITHUB_OUTPUT_EOF` as input to `$GITHUB_OUTPUT`. Then in the third line we actually add `GITHUB_OUTPUT_EOF` manually to signal that the input is over. The only caveat is that you have to make sure that `GITHUB_OUTPUT_EOF` does not appear in `$ERROR_MSG` (which I think is pretty unlikely to happen).

CharlesDuboisSAP: Why do we need to do that? What made it fail?

CharlesDuboisSAP (approved):

vladimir-a-sap (approved):
LGTM

PR #945, opened 2026-07-14
fix: [OpenAI] bug `.withHeader` deletes context from `OpenAiClient`

n-o-u-r-h-a-n:
## Context

AI/ai-sdk-java-backlog#415.

Singular method `.withHeader` deletes context from `OpenAiClient` so was fixed and also created `.withHeaders` which accepts multiple custom headers.

### Feature scope:
 
- [x] bug in `.withHeader` fixed to include systemPrompt.
- [x] implemented a `OpenAiClient.withHeaders(Map<String, String> headers)` method
- [x] Added unit tests to both changes 

## Definition of Done

- [x] Release notes updated

### DIFF
--- a/foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
+++ b/foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
@@ -507,4 +507,45 @@ class OpenAiClientTest extends BaseOpenAiClientTest {
             .withHeader("Header-For-Both", equalTo("value"))
             .withHeader("foot", equalTo("baz")));
   }
+
+  @Test
+  @SuppressWarnings("deprecation")
+  void testWithHeaderPreservesSystemPrompt() {
+    stubForChatCompletion();
+
+    client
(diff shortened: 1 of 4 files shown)


foundation-models/openai/src/test/java/com/sap/ai/sdk/foundationmodels/openai/OpenAiClientTest.java
```diff
@@ -507,4 +507,45 @@ void testCustomHeaders() {
             .withHeader("Header-For-Both", equalTo("value"))
             .withHeader("foot", equalTo("baz")));
   }
+
+  @Test
+  @SuppressWarnings("deprecation")
```
CharlesDuboisSAP, line 512: What is depercated?

n-o-u-r-h-a-n: Chat completion method which gets the prompt as an input, but it was the only one which could work for my test that does not ignore the systemPrompt field as chat completion method which receives OpenAiChatCompletionRequest ignores the client-level systemPrompt.

CharlesDuboisSAP (approved):

PR #950, opened 2026-07-16
fix: e2e logging issue

vladimir-a-sap:
[task](https://github.com/SAP/ai-sdk-java-backlog/issues/417)

## Context

Fix slack notification

### Feature scope:
 
- Fixes slack notification with failed e2e tests names

## Definition of Done

- [X] Functionality scope stated & covered
- [X] Tests cover the scope above (manually tested)

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -52,7 +52,7 @@ jobs:
           MVN_ARGS="${{ env.MVN_MULTI_THREADED_ARGS }} surefire:test -pl :spring-app -DskipTests=false"
           mvn $MVN_ARGS "-Daicore.landscape=${{ matrix.environment }}" 2>&1 | tee mvn_output.log # tee writes to both the console and a file
 
-          ERROR_MSG=$(grep -E '^\[ERROR\] +[A-Za-z][A-Za-z0-9_]*\.[a-zA-Z][a-zA-Z0-9_]*:[0-9]+' mvn_output.log | sed $'s/\\[ERROR\\] */\\\n 🔴/' || true)
+          ERROR_MSG=$(grep -E '^\[ERROR\] +[A-Za-z][A-Za-z0-9_]*\.[a-zA-Z][a-zA-Z0-9_]*:[0-9]+' mvn_output.log | sed $'s/\\[ERROR\\] */\\\n 🔴 /' || true)
           echo "$ERROR_MSG"
           {
             echo "error_message<<EOF"
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -116,5 +116,5 @@ jobs:
                     text: "⚠️ End-to-end tests failed! 😬 Please inspect & fix by clicking <https://github.com/SAP/ai-sdk-java/actions/runs/${{ github.run_id }}|here>"
               - type: "section"
                 text:
-                  type: "plain_text"
+                  type: "mrkdwn"
                   text: "${{ steps.run_tests.outputs.error_message }} "
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -103,7 +103,7 @@ jobs:
         run: wget -qO- -S localhost:8080
 
       - name: "Slack Notification"
-        #if: failure()
+        if: failure()
         uses: slackapi/slack-github-action@v3.0.3
         with:
           webhook: ${{ secrets.SLACK_WEBHOOK }}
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-test.yaml
```diff
@@ -113,5 +116,5 @@ jobs:
                     text: "⚠️ End-to-end tests failed! 😬 Please inspect & fix by clicking <https://github.com/SAP/ai-sdk-java/actions/runs/${{ github.run_id }}|here>"
               - type: "section"
                 text:
-                  type: "plain_text"
+                  type: "mrkdwn"
```
CharlesDuboisSAP, line 119: This doesn't change anything right?

sample-code/spring-app/pom.xml
```diff
@@ -312,6 +312,7 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
+          <forkCount>4C</forkCount>
```
CharlesDuboisSAP: Would like the team's opinion on this, it would scramble the logs instead of having them ordered. On the other hand, we could have a maven wide config to run all modules run in parallel to speed up everything

.github/workflows/e2e-test.yaml
```diff
@@ -50,12 +50,15 @@ jobs:
           fi
           
           MVN_ARGS="${{ env.MVN_MULTI_THREADED_ARGS }} surefire:test -pl :spring-app -DskipTests=false"
-          mvn $MVN_ARGS "-Daicore.landscape=${{ matrix.environment }}" | tee mvn_output.log # tee writes to both the console and a file
+          mvn $MVN_ARGS "-Daicore.landscape=${{ matrix.environment }}" 2>&1 | tee mvn_output.log # tee writes to both the console and a file
```
CharlesDuboisSAP, line 53: What happens without `2>&1`?

vladimir-a-sap: If some tests fail, error logs might reside elsewhere upper in the logs and it could be difficult to match errors with failed test names and identify causes.

The reason for this is that std output is normally buffered, std error output is not (immediate output vs potentially delayed). 

If error output is not redirected to the same output as std, read from logs output chronology might be incorrect.

`2>&1` statement redirects output 2 (standard error) to output 1 (std)

sample-code/spring-app/pom.xml
Jonas-Isr: I don't see the scrambling of logs as an issue since for actual debugging of the test we can always run them locally. What I think is important from the logs is *that* something breaks and *what*, but necessary *why*. The why I usually debug locally anyways.

One concern could be that we have some e2e tests that create/delete things on the backend (for example in `PromptRegistryTest`). I am not sure if it is safe to run these in parallel. But if we checked that, I would be in favour of speeding things up.

.github/workflows/e2e-test.yaml
Jonas-Isr: I think plaint_text does not render the emojies correctly.

sample-code/spring-app/pom.xml
vladimir-a-sap: Would be open to discuss it.

I was thinking to make all phases parallel, but came to conclusion that it would be detrimental for build process at least in certain cases:

- our CI runner apparently has just 1 core
- build process utilizes this 1 core fully (parallelization here would only slow down build adding thread sync overhead)
- e2e tests underutilize the core, waiting on IO ~90%+ of the time and idling the CPU
- added this thing while debugging and found that it cuts e2e test time by approx 4 minutes

Jonas-Isr (commented):
Changes in e2e-test workflow lgtm.

I would approve once we checked whether parallelisation is safe for our e2e tests.

.github/workflows/e2e-test.yaml
vladimir-a-sap: I wasn't able to make plugin correctly handle `/n` (newline) symbols with `plain_text`, may be I was missing something, but with this combination (`mrkdwn` and the current call args) it works

sample-code/spring-app/pom.xml
vladimir-a-sap: removed, given there are doubts

.github/workflows/e2e-test.yaml
CharlesDuboisSAP: You say « might ». Why not test it to see the difference, maybe it’s not needed

vladimir-a-sap: This issue is of probabilistic nature. It will **certainly** occur in general, but with a chance for a particular test/run, therefore "might"

vladimir-a-sap: if you don't like it or not agree that it is a right solution I'm happy to revert it though

Jonas-Isr: I would say it makes sense to try this out. If we experience any degradation in the logs we are always free re-evaluate. In the end, this is a pipeline we use internally and not productive code, so we can experiment (and potentially revert).

Jonas-Isr (approved):
Looks good from my side

Of course, let's only merge if all reviewers agree :)

CharlesDuboisSAP (approved):

PR #955, opened 2026-07-22
feat: support prompt caching and add tests

vladimir-a-sap:
## Context

[AI/ai-sdk-java-backlog#413](https://github.com/SAP/ai-sdk-java-backlog/issues/413)

Adds prompt caching support (UserMessage, SystemMessage, ToolMessage) for new Claude models.

Documentation pr: https://github.com/SAP/ai-sdk/pull/644

## Definition of Done

- [x] Release notes updated


docs/release_notes.md
```diff
@@ -22,6 +22,11 @@
 - [OpenAI] You can now add multiple custom headers to an `OpenAiClient` at once via `.withHeaders()`.
 - [Orchestration] Added `GEMINI_3_1_PRO_PREVIEW_EA` and `MISTRAL_MEDIUM` to model list in `OrchestrationAiModel`.
 - [Orchestration] Reasoning content is now supported for users through `OrchestrationChatResponse.getReasoningText()` and `OrchestrationChatCompletionDelta.getDeltaReasoningText()` to see the thinking processes when using the reasoning models.
+- [Orchestration] Added support for prompt caching for new Claude models. The following messages are cacheable:
+  - `SystemMessage`
+  - `UserMessage`
+  - `ToolMessage`
+  - refer to `com.sap.ai.sdk.orchestration.PromptCachingConfig.PROMPT_CACHING_SUPPORT` for the models supporting caching
```
Jonas-Isr: (Minor)

```suggestion
- [Orchestration] Added support for prompt caching for new Claude models. The following messages are cacheable:
  - `SystemMessage`,
  - `UserMessage`,
  - `ToolMessage`.
  - Refer to `com.sap.ai.sdk.orchestration.PromptCachingConfig.PROMPT_CACHING_SUPPORT` for the models supporting caching.
```

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ToolMessage.java
Jonas-Isr, line 75: (Major)

Shouldn't this also include the cacheControl?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ConfigToRequestTransformer.java
```diff
@@ -98,6 +110,75 @@ static PromptTemplatingModuleConfigPrompt toTemplateModuleConfig(
...
+            }
+          }
+        }
+        contentItems.add(contentItem);
+      }
+      if (message instanceof UserMessage) {
+        message = new UserMessage(new MessageContent(contentItems));
+      } else if (message instanceof SystemMessage) {
+        message = new SystemMessage(new MessageContent(contentItems));
+      }
```
Jonas-Isr, line 174: (Major/Question)

Is it correct that this does not check for `ToolMessage` here?

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ConfigToRequestTransformer.java
```diff
@@ -98,6 +110,75 @@ static PromptTemplatingModuleConfigPrompt toTemplateModuleConfig(
...
+        var contentItem = message.content().items().get(j);
+        if (contentItem instanceof TextItem textItem) {
+          final var existingCacheControl = textItem.getCacheControl();
+          if (existingCacheControl != null) {
+            if (remainingCacheableCheckpoints <= 0 || !messageSupportsCache) {
+              // skipping text control if there is already too many cache checkpoints
+              contentItem = new TextItem(textItem.text());
+            } else if (cachingConfig.getMinTokensPerCheckpoint()
+                > textItem.getText().chars().filter(c -> c == ' ').count() + 1) {
+              // to few tokens to cache, skipping to respect model limitation
```
Jonas-Isr:
```suggestion
              // too few tokens to cache, skipping to respect model limitation
```

orchestration/src/main/java/com/sap/ai/sdk/orchestration/PromptCachingConfig.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Pattern;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import lombok.Getter;
+
+/** Describes supported caching properties of a model */
+@Getter
+public final class PromptCachingConfig {
```
Jonas-Isr: (Minor)

I think this class could be a `@Value`. That would also add equals/hash code.

orchestration/src/main/java/com/sap/ai/sdk/orchestration/SystemMessage.java
```diff
@@ -48,15 +64,53 @@ public SystemMessage(@Nonnull final String message) {
...
+  /**
+   * Add text to the message
+   *
+   * @since 1.23.0
+   * @param message the text to add
+   * @param cacheControl optional cache checkpoint configuration
+   * @return the new message
+   */
+  @SuppressWarnings(
+      "PMD.PublicApiExposesModelType") // only exposes obj from the same package, already a dto
```
Jonas-Isr: (Question)

Is this necessary because PMD confuses the two CacheControl classes?

docs/release_notes.md
vladimir-a-sap: corrected as you proposed

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ConfigToRequestTransformer.java
vladimir-a-sap: corrected

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ConfigToRequestTransformer.java
vladimir-a-sap: thanks, that was a missing part, fixed!

orchestration/src/main/java/com/sap/ai/sdk/orchestration/PromptCachingConfig.java
vladimir-a-sap: converted to `@Value` as you proposed

orchestration/src/main/java/com/sap/ai/sdk/orchestration/SystemMessage.java
vladimir-a-sap: It is necessary, but I do not know if it is because PMD doesn't consider packages while comparing classes or for another reason (e.g. ruleset limitations)

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ToolMessage.java
vladimir-a-sap: thanks for catching that! corrected the method

orchestration/src/main/java/com/sap/ai/sdk/orchestration/SystemMessage.java
Jonas-Isr: After checking it looks like this is indeed a confusion on the side of the PMD rule. When we wrote that rule we simply compared the simple names of the classes imported, not their fully qualified names. I proposed a change to the comment next to the `@SuppressWarning` to make this clear (see comment below).

Jonas-Isr:
```suggestion
      "PMD.PublicApiExposesModelType") // false positive: the two CacheControl classes are mixed up by the rule
```

orchestration/src/main/java/com/sap/ai/sdk/orchestration/PromptCachingConfig.java
```diff
@@ -0,0 +1,108 @@
...
+          });
+
+  /** Caching checkpoint can only be made for so few prompt input tokens and not fewer */
+  int minTokensPerCheckpoint;
+
+  /** Only up to this number of caching points can be created per request */
+  int maxCheckpointsPerRequest;
+
+  /** Pattern of supported TTL values, which can be passed */
+  Pattern supportedTTLValues;
```
Jonas-Isr: (Minor)

After looking at this again, I am not sure if it is the best approach to expose this in this form. Aren't we basically exposing internal details by giving access to the exact pattern? Maybe a method like `boolean isTTLSupported(String ttl)` would be better where we check the pattern on our side and just give a boolean response?

Jonas-Isr (commented):
Mostly LGTM :)

orchestration/src/main/java/com/sap/ai/sdk/orchestration/PromptCachingConfig.java
vladimir-a-sap: addressed. this class by intention is internal only, I also renamed it to `ModelPromptCachingSupport` and made package local to exclude any confusion

orchestration/src/main/java/com/sap/ai/sdk/orchestration/SystemMessage.java
vladimir-a-sap: accepted locally

vladimir-a-sap: accepted (locally) you suggestion and changed other same comments in two more places

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ModelPromptCachingSupport.java
```diff
@@ -77,7 +76,7 @@ private PromptCachingConfig(
    * @return model prompt caching configuration
    */
   @Nonnull
-  public static PromptCachingConfig forModel(@Nullable final String modelName) {
+  public static ModelPromptCachingSupport forModel(@Nullable final String modelName) {
```
Jonas-Isr: (Minor)

If the class is package private now, I would prefer to not have the `public` modifier on these methods. Even though they don't do anything they could be confusing in the future.

docs/release_notes.md
```diff
@@ -29,7 +29,7 @@
   - `SystemMessage`,
   - `UserMessage`,
   - `ToolMessage`.
-  - Refer to `com.sap.ai.sdk.orchestration.PromptCachingConfig.PROMPT_CACHING_SUPPORT` for the models supporting caching.
+  - Refer to `com.sap.ai.sdk.orchestration.ModelPromptCachingSupport.PROMPT_CACHING_SUPPORT` for the models supporting caching.
```
Jonas-Isr: (Question)

Does it make sense to refer to `ModelPromptCachingSupport.PROMPT_CACHING_SUPPORT` here even though that is a private field in a package private class? Are you expecting users to look at the source code? Then it is maybe more user friendly to just have this list in our actual docs somewhere? (That would have the downside of being 2 lists to maintain, though. Not sure if that is better.)

vladimir-a-sap: removed reference here. supported caching for each model is already specified in the docs PR

orchestration/src/main/java/com/sap/ai/sdk/orchestration/ModelPromptCachingSupport.java
vladimir-a-sap: made package-private

Jonas-Isr (approved):
LGTM

PR #991, opened 2026-08-21
chore: [DevOps] Revert temporary switch to NVD datafeed

Jonas-Isr:
## Context

Earlier this year we [had to switch to the NVD datafeed](https://github.com/SAP/ai-sdk-java/pull/921) because their API became unusable. That switch was planned as a temporary solution. This PR switches back to using the API.

Successful run [here](https://github.com/SAP/ai-sdk-java/actions/runs/32487861759).


vladimir-a-sap (approved):
discussed the context, LGTM given the original API now works again

PR #992, opened 2026-08-21
chore: [DevOps] Ignore Spring AI v2 update

Jonas-Isr:
## Context

This PR lets the dependabot check ignore Spring AI v2 for the moment until we can actually migrate to it.


vladimir-a-sap (approved):
approved, we agreed to address this as a separate task as it is a major chunk of work, not just a tiny fix

PR #1001, opened 2026-08-26
fix: [DevOps] NVD scan remove branch specific caches and increase timeout

CharlesDuboisSAP:
## Context

- https://github.com/SAP/cloud-sdk-java-backlog/issues/448

Please provide a short description of what your change does and why it is needed.

### Feature scope:
 
- [x] The only cache used will be main on all branches, only main branches run will delete and re-update the cache.
- [x] Increase timeout to 1 hour


vladimir-a-sap (approved):
1h connection timeout looks weird, but let's see if it helps to fix the issue

PR #1002, opened 2026-08-26
fix: Fix failed e2e tests slack notifications

vladimir-a-sap:
## Context

Fixed regexp to support `test timeout` logs which are (pattern-wise) different from typical test failures


Jonas-Isr (approved):
LGTM

PR #1016, opened 2026-08-31
fix: possible race condition sending wss packets

vladimir-a-sap:
**Context / Cause**
- Java WSS.send(...) does not actually send data synchronously (failed expectation)
- Reordered packets (e.g. sending something after model response has been requested) interrupts ongoing model response
- Test waits until timeout and fails

**Fix**
- `ws.send(...)` replaced with `ws.send(...).join()` in client.send(...)
- fix confirmed after [100 successful test runs in CI](https://github.com/SAP/ai-sdk-java/actions/runs/33415423673/job/99564795275)


docs/release_notes.md
```diff
@@ -20,4 +20,4 @@
 
 ### 🐛 Fixed Issues
 
--
+- [Foundation models] Fixed possible race condition in Realtime API implementation
```
CharlesDuboisSAP:
```suggestion
- [Realtime] Fixed possible race condition in Realtime API implementation
```
http://sap.github.io/ai-sdk/docs/java/release-notes#1230---august-20-2026

CharlesDuboisSAP (commented):
Would you need to change the Javadoc to indicate that the API is blocking or was this already the expectation?

vladimir-a-sap:
> Would you need to change the Javadoc to indicate that the API is blocking or was this already the expectation?

no, the exposed api signatures are all sync, no changes there or in the docs are needed

Jonas-Isr (approved):

PR #1018, opened 2026-09-01
fix: rpt e2e tests

vladimir-a-sap:
Implements behavior to avoid user-facing breaking changes between rpt 1.5.0 and 1.6.0 specs


Jonas-Isr (approved):

vladimir-a-sap:
checked E2E tests locally prior to merging, all green

PR #1020, opened 2026-09-01
fix: Handle contextMode in RptClient

Jonas-Isr:
## Context

I added some logic to `RptClient` to set `contextMode` to null whenever a version before 1.6 is used.

E2E tests run green for me on this locally.

### DIFF
--- a/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/RptService.java
+++ b/sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/RptService.java
@@ -53,26 +53,25 @@ public class RptService {
             Map.of(
                 "PRODUCT", RowsInnerValue.create("Couch"),
                 "PRICE", RowsInnerValue.create(BigDecimal.valueOf(999.99)),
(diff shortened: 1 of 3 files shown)


sample-code/spring-app/src/main/java/com/sap/ai/sdk/app/services/RptService.java
```diff
@@ -53,26 +53,25 @@ public PredictResponsePayload predict() {
             Map.of(
                 "PRODUCT", RowsInnerValue.create("Couch"),
                 "PRICE", RowsInnerValue.create(BigDecimal.valueOf(999.99)),
-                "ORDERDATE", RowsInnerValue.create("2025-11-28"),
+                "ORDERDATE", RowsInnerValue.create("28-11-2025"),
```
Jonas-Isr: I think these where not needed.

vladimir-a-sap (dismissed):
discussed the tradeoffs, seems like a balanced approach to avoid breaking changes for our users

vladimir-a-sap (approved):

PR #1028, opened 2026-09-04
feat: [TabularOrchestration] Added Tabular orchestration module

CharlesDuboisSAP:
## Context

- https://github.com/SAP/ai-sdk-java-backlog/issues/422

New AI Core service

## Definition of Done

  - https://github.com/SAP/ai-sdk/pull/706
- [x] Release notes updated


.github/workflows/spec-update.yaml
```diff
@@ -171,6 +173,16 @@ jobs:
               FILE_PATH='core-services/batch/src/main/resources/spec/batch-service.yaml'
               MODULE_PATH='core-services/batch'
               ;;
+            tabular-orchestration) #https://github.tools.sap/AI/context-registry/blob/main/docs/public/business_api_hub/api_hub_merged_spec.yaml
+              API_URL="$API_BASE_URL/AI/context-registry/contents/docs/public/business_api_hub/api_hub_merged_spec.yaml?ref=$REF"
+              FILE_PATH='foundation-models/tabular-orchestration/src/main/resources/spec/tabular-orchestration.yaml'
```
Jonas-Isr: Here and below.

```suggestion
              FILE_PATH='core-services/tabular-orchestration/src/main/resources/spec/tabular-orchestration.yaml'
```

.pipeline/spotbugs-exclusions.xml
```diff
@@ -11,6 +11,7 @@
       <Package name="com.sap.ai.sdk.foundationmodels.rpt.generated.model"/>
       <Package name="com.sap.ai.sdk.prompt.registry.model"/>
       <Package name="com.sap.ai.sdk.prompt.registry.client"/>
+      <Package name="com.sap.ai.sdk.foundationmodels.tabular.generated"/>
```
Jonas-Isr:
```suggestion
      <Package name="com.sap.ai.sdk.tabular.generated"/>
```

core-services/tabular-orchestration/src/main/java/com/sap/ai/sdk/tabular/TabularClient.java
```diff
@@ -0,0 +1,139 @@
...
+import lombok.val;
+
+/**
+ * Service class for the Tabular Orchestration APIs.
+ *
+ * @since 1.26.0
+ */
+@RequiredArgsConstructor(access = AccessLevel.PUBLIC)
+@Getter(value = AccessLevel.PROTECTED)
+public class TabularClient {
```
Jonas-Isr, line 32: (Question)

Should we mark this `@Beta`?

Jonas-Isr (commented):
New convenience API looks good! Besides that had only smaller comments.

Once the spec seems stable we can release this in my opinion :)

CharlesDuboisSAP:
> New convenience API looks good! Besides that had only smaller comments.
> 
> Once the spec seems stable we can release this in my opinion :)

@Jonas-Isr I think we could already merge it as beta.
You should also look into the resource group for predict API before merging

core-services/tabular-orchestration/src/main/java/com/sap/ai/sdk/tabular/TabularClient.java
```diff
@@ -0,0 +1,153 @@
...
+
+  /**
+   * Get the Predict API.
+   *
+   * @return The Predict API.
+   */
+  @Nonnull
+  public PredictApi predict() {
+    return predict("default");
+  }
```
Jonas-Isr, line 96: (Minor) 


```suggestion
  /**
   * Get the Predict API. This uses the resource group "default".
   *
   * @return The Predict API.
   */
  @Nonnull
  public PredictApi predict() {
    return predict("default");
  }
```

core-services/tabular-orchestration/src/main/java/com/sap/ai/sdk/tabular/TabularClient.java
```diff
@@ -0,0 +1,153 @@
...
+   */
+  @Nonnull
+  public PredictApi predict(@Nonnull final String resourceGroup) {
+    final HttpDestination inferenceDestination =
+        predictDestinationOverride != null
+            ? predictDestinationOverride
+            : getService().getInferenceDestination(resourceGroup).forScenario(PREDICT_SCENARIO_ID);
+    val destination =
+        DefaultHttpDestination.fromDestination(inferenceDestination).headers(customHeaders).build();
+    return new PredictApi(ApiClient.create(destination));
```
Jonas-Isr, line 113: (Question)

This means, the resource group is set once for the PredictApi object. I assume this is because the two specs are different, right? 

That's a bit annoying because of the inconsistency in usage but I like the way the public API looks, given this inconsistency is nothing we can change.

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/TabularTest.java
```diff
@@ -0,0 +1,134 @@
...
+import com.sap.ai.sdk.app.services.TabularService.DataDestinationService;
+import com.sap.ai.sdk.app.services.TabularService.PredictionService;
+import com.sap.ai.sdk.app.services.TabularService.ScenarioConfigurationService;
+import com.sap.ai.sdk.tabular.generated.orchestration.model.HDLDataDestinationGetResponse;
+import lombok.extern.slf4j.Slf4j;
+import lombok.val;
+import org.junit.jupiter.api.Test;
+
+@Slf4j
+class TabularTest {
```
Jonas-Isr, line 16: (Minor)

Can we rename this to `TabularOrchestrationTest`? Since also the `RptTest` class is somewhat about tabular stuff.

Jonas-Isr:
After discussion, I would rather wait until the Predict spec is stabilised and working properly.

core-services/tabular-orchestration/src/main/java/com/sap/ai/sdk/tabular/TabularClient.java
CharlesDuboisSAP: No, it's because if I let the spec define the resource group, then the code generates like so:
```java
 new TabularClient().predict().predict(resourceGroup, request);
```
The first `predict` gets the deployment on the default resource group, then the request with custom resource group will fail because it uses the default deployment.

Now if you set the resource group for fetching the deployment:
```java
 new TabularClient().predict(resourceGroup).predict(resourceGroup, request);
```
Then the resource group propagates to the client and it's set twice, again the request fails because AI Core doesn't like it.

That's why I chose setting it once for the client and let it propagate to the request which is inconsistant:
```java
 new TabularClient().predict(resourceGroup).predict(request);
```

sample-code/spring-app/src/test/java/com/sap/ai/sdk/app/controllers/TabularTest.java
CharlesDuboisSAP: Just the test class or everything?

ZhongpinWang (dismissed):
please consider separate `context-registry` and `tabular-ai-orchestration` into two different modules. They are two independent services.

Also rename the spec names, and folders. Context registry is not tabular orchestration. Tabular AI orchestration has one endpoint called /predict but the service is not called tabular predict.

core-services/tabular-ai-orchestration/src/main/resources/spec/tabular-ai-orchestration.yaml
```diff
@@ -0,0 +1,773 @@
+openapi: 3.0.3
+info:
+  title: Tabular AI Orchestration
```
CharlesDuboisSAP, line 3: The specs should be updated.

core-services/tabular-ai-orchestration/src/test/java/com/sap/ai/sdk/tabular/orchestration/TabularOrchestrationUnitTest.java
```diff
@@ -0,0 +1,131 @@
...
+    assertThat(thirdPrediction.get(0).get("prediction")).isEqualTo("Furniture");
+
+    assertThat(response.getStatus().getCode()).isEqualTo(0);
+    assertThat(response.getStatus().getMessage()).isEqualTo("ok");
+    assertThat(response.getAdditionalInformation()).isEmpty();
+  }
+
+  //  @Test
+  //  void testCustomHeaders() {
+  //    WireMock.stubFor(
```
CharlesDuboisSAP, line 113: This test should be created using e2e test payloads

core-services/tabular-ai-orchestration/src/main/resources/spec/tabular-ai-orchestration.yaml
ZhongpinWang: Please let me know if you see anything problematic in the spec. I will ask them to adjust but may take a while.

PR #1044, opened 2026-09-11
chore: [DevOps] Stagger different e2e test runs by one hour

CharlesDuboisSAP:
## Context

- https://github.com/SAP/ai-sdk-java-backlog/issues/429

Reduces flakiness by making both e2e test branches be run and different hours

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -2,7 +2,8 @@ name: "End-to-end Tests"
 on:
   workflow_dispatch:
   schedule:
-    - cron: 15 2 * * MON-FRI
+    - cron: 15 1 * * MON-FRI # main, runs one hour before v2
+    - cron: 15 2 * * MON-FRI # v2
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-test.yaml
```diff
@@ -12,6 +13,10 @@ jobs:
   end-to-end-tests:
     permissions:
       contents: read
+    if: |
+      github.event_name != 'schedule' ||
```
vladimir-a-sap, line 17: shouldn't the logic be inverted, `github.event_name == 'schedule` ?

I recall in other places we use `if: failure()` for slack notification, that seem to imply that `true` is expected

vladimir-a-sap (approved):
good to go if the `if` thing is confirmed to be correct

.github/workflows/e2e-test.yaml
vladimir-a-sap: fyi @CharlesDuboisSAP

CharlesDuboisSAP: We want the workflow to still run with ‘dispatch’

PR #1048, opened 2026-09-15
fix: [DevOps] Fix e2e-test set up

Jonas-Isr:
## Context

This PR fixes a problem [introduced in the most recent e2e test change](https://github.com/SAP/ai-sdk-java/pull/1044). The problem is that `matrix.branch` is not available at job level but only on steps level. This led to github not parsing the yaml file.

I pulled the if into the steps level. It is less concise but works. Happy for other options.

[Successful run here](https://github.com/SAP/ai-sdk-java/actions/runs/34953318659).


.github/workflows/e2e-test.yaml
```diff
@@ -31,24 +27,44 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
+      - name: "Check schedule filter"
```
vladimir-a-sap, line 30: (non-blocking proposal):
- instead of complex scheduling and filtering hacks, revert back to the e2e job taking 'branch' as an argument and 'slack_notification_on_failure' flag with default branch being 'main' and 'slack_notification_on_failure' being 'false' by default.
- add two trivial scheduled workflows with one single step each to trigger this workflow with right arguments

vladimir-a-sap (approved):
- acceptable as a temporary fix to be reverted after we stop supporting v1
- a potentially cleaner and less brittle solution proposed in the comment

.github/workflows/e2e-test.yaml
Jonas-Isr: Yes I thought about that but did not want to add two additional workflows just for this temporary fix...

I will merge the PR since you agreed to the current solution as well :)

